PR #3, opened 2023-04-05
[WIP] Move bindings, part 3

mitschabaude:


mitschabaude:
this starts something valuable (making `Proof` a normal `Provable<T>`, not some special case), but there's not enough meat on this PR to justify keeping it open

PR #6, opened 2023-04-18
Clean up previous proof arguments handling in Pickles/SnarkyJS

mitschabaude:
this simplifies the Pickles bindings by handling the public inputs and outputs of proofs on the JS side, instead of in OCaml -- consistent with how other private inputs are handled


ocaml/lib/snarky_js_bindings_lib.ml
```diff
@@ -2034,64 +2071,16 @@ module Choices = struct
           ; main =
               (fun { public_input } ->
                 dummy_constraints () ;
-                (* TODO: Push this down into SnarkyJS so that it controls the
-                   public inputs of prev proofs, and we can delete this
-                   annoying logic.
-                *)
```
mitschabaude, line 2040: this TODO is resolved now

Trivo25 (approved):

PR #8, opened 2023-05-02
Support custom gates in Wasm bindings

rbonichon:
This adds the necessary code to support custom gates in wasm bindings


mitschabaude (changes requested):
I have one important comment!

kimchi/wasm/src/lib.rs
```diff
@@ -73,7 +74,7 @@ pub fn wait_until_non_zero(ptr: *const u32) -> u32 {
     unreachable!();
 }
 
-pub mod rayon;
+pub use wasm_bindgen_rayon::init_thread_pool;
```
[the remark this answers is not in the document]
rbonichon: indeed. let me fix this.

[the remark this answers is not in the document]
rbonichon: fixed!

mitschabaude (approved):

mrmr1993 (changes requested):
Please don't mix in unrelated changes.

It looks like the JS side isn't complete, since it never calls the newly added accessors, or uses the new arguments. If you search for `emul_comm` in the JS code, you should find the relevant places.

rbonichon:
> Please don't mix in unrelated changes.

Clippy suggestions and/or missing functions in JS bindings will be moved to other PRs

kimchi/js/bindings.js
```diff
@@ -1424,6 +1447,12 @@ var caml_plonk_verification_evals_to_rust = function (x, klass, poly_comm_class,
         return caml_poly_comm_to_rust_poly_comm(poly_comm, poly_comm_class, mk_affine);
     };
 
+    var convert_opt = function (poly_comm_opt) {
```
mrmr1993, line 1450: I don't think this does what we want. The JSOO representation of an OCaml `option` is `0` for `None` or `[0, foo]` for `Some foo`. You can see similar being done elsewhere in this file

rbonichon: Doesn't it? Don't we want to map `None` to `undefined` and `Some x` to `caml_poly_comm_to_rust_poly_comm(x, ....)`  in this `to_rust` function ?

rbonichon: AFAICT, the code here is very similar to `caml_poly_comm_to_rust_poly_comm` where we have the following snippet
```
    var caml_unshifted = poly_comm[1];
    var caml_shifted = poly_comm[2];
    var rust_shifted = undefined;
    if (caml_shifted !== 0) {
        rust_shifted = rust_affine_of_caml_affine(caml_shifted[1], mk_affine);
    }
```

mrmr1993 (approved):

PR #9, opened 2023-05-03
Move Provable (formerly Circuit) to JS

mitschabaude:
companion of https://github.com/o1-labs/snarkyjs/pull/889

this adds a new exported module `snarky` to the OCaml side of snarkyjs, which is much simpler and closer to the OCaml API than the previous `circuit` export. `circuit` is deleted, along with a few other unused parts of the bindings_lib.

the github PR view makes the changes hard to see -- I recommend reviewing the implementation of the new module here:

https://github.com/o1-labs/snarkyjs-bindings/blob/8b40c9a3eb82f9852e913da390284643f65d2bff/ocaml/lib/snarky_js_bindings_lib.ml#L1244-L1355


Trivo25:
Should we generally try to (maybe not now, in a follow up PR) cleanup this massive `snarky_js_bindings_lib.ml` OCaml file into its own subsections, e.g. `snarky.ml`, `pickles.ml`, .. or would that be too annoying?

ocaml/lib/snarky_js_bindings_lib.ml
```diff
@@ -1248,563 +1219,163 @@ let poseidon =
...
-        (fun x -> main x)
-    in
-    let kp = Impl.Keypair.generate ~prev_challenges:0 cs in
-    new%js keypair_constr kp
+  let run_unchecked (f : unit -> unit) =
+    try
+      Impl.run_and_check_exn (fun () ->
+          Snarky_backendless.Snark0.set_eval_constraints false ;
+          f () ;
+          Snarky_backendless.Snark0.set_eval_constraints true ;
```
Trivo25, line 1272: is it because of the global state that we have to switch between these modes? :D

ocaml/lib/snarky_js_bindings_lib.ml
```diff
@@ -1248,563 +1219,163 @@ let poseidon =
...
+        main'
     end
 
-  let prove (type w p) (c : (w, p) Circuit_main.t) (priv : w) (pub : p) kp :
-      proof_class Js.t =
-    let main, input_typ = main_and_input c in
-    let pk = Keypair.pk kp in
-    let p =
-      Impl.generate_witness_conv ~return_typ:Snark_params.Tick.Typ.unit
+    let compile main public_input_size =
```
Trivo25, line 1298: when we talk about compile here, what does this mean? Is this Kimchi compiling or translating the circuit into a constraint system?

ocaml/lib/snarky_js_bindings_lib.ml
mitschabaude: yes, in this function we want checks like `x.assertEquals(y)` to not throw (because we want to be able to use dummy data), and therefore we have to toggle this global flag

ocaml/lib/snarky_js_bindings_lib.ml
mitschabaude: yes, both steps are here:
* `Impl.constraint_system` is running the function and recording constraints & variables
* `Impl.Keypair.generate` is Kimchi creating the prover and verification keys from the constraint system

PR #12, opened 2023-05-08
Move Field to JS

mitschabaude:
This exposes snarky methods on field elements, in service of https://github.com/o1-labs/snarkyjs/pull/902


Trivo25 (approved):

ocaml/lib/snarky_js_bindings_lib.ml
```diff
@@ -1288,6 +1288,66 @@ module Snarky = struct
...
+    let from_bits bits = Array.to_list bits |> Field.project
+
+    (** returns x truncated to the lowest [length] bits
+       => can be used to assert that x fits in [length] bits.
+
+       more efficient than [to_bits] because it uses the [EC_endoscalar] gate;
+       does 16 bits per row (vs 1 bits per row that you can do with generic gates).
+       [length] has to be a multiple of 16
+    *)
+    let truncate_to_bits (length : int) x =
```
bkase, line 1336: can we make this parameter `length_div_16` instead and multiply by 16 in the body of the function -- then the function is total and the caller can prove that it's a multiple of 16 by dividing it by 16 before passing it in.

Consider changing the name to `truncate_to_bits16` also

bkase (changes requested):

ocaml/lib/snarky_js_bindings_lib.ml
mitschabaude: the input length is a plain integer, which will define the circuit statically, so the caller can't / doesn't have to prove anything about it.

do you still think we should change the signature?
I'm going to change it here but not in the user-facing SnarkyJS method for now (but will throw a quick error there)

bkase (approved):

PR #14, opened 2023-05-08
Add missing JS binding functions

rbonichon:
- group *_example_* custom gates test functions together at the end
- provide simple int64_of_int32 missing conversion function
- add caml_fq_srs_add_lagrange_basis, whichs an updating function (no return value)


kimchi/js/bindings.js
```diff
@@ -2259,12 +2265,60 @@ function caml_pasta_fq_poseidon_block_cipher(_fake_params, fq_vector) {
...
-}
 
 // Provides: prover_to_json
 // Requires: plonk_wasm
 var prover_to_json = plonk_wasm.prover_to_json;
+
+// Provides: integers_uint64_of_uint32
+function integers_uint64_of_uint32(i) {
+    // Same as integers_uint64_of_int
+    return new new UInt64(caml_int64_of_int32(i));
```
mitschabaude, line 2276: there are two `new`s here, and `UInt64` as well as `caml_int64_of_int32` seem to be missing as a dependency? You annotate a dependency with `// Requires: ...`

mitschabaude (commented):
thanks for adding these!

kimchi/js/bindings.js
rbonichon: Done

rbonichon: This should be fixed.

mitschabaude:
This PR should specify what Mina / snarkyjs PR it is compatible with

mitschabaude (approved):
I added a snarkyjs PR which includes these changes, to test them in CI: https://github.com/o1-labs/snarkyjs/pull/1001

querolita (approved):

dannywillems:
LGTM, waiting for the CI

PR #19, opened 2023-05-10
fix Error: Cannot find module on Windows OS

wizicer:
This patch can fix: https://github.com/o1-labs/snarkyjs/issues/896

### DIFF
--- a/js/node/node-backend.js
+++ b/js/node/node-backend.js
@@ -2,7 +2,7 @@ import { isMainThread, parentPort, workerData, Worker } from 'worker_threads';
 import os from 'os';
 import wasm_ from '../../../_node_bindings/plonk_wasm.cjs';
 import { fileURLToPath } from 'url';
-let filename = fileURLToPath(import.meta.url);
+let filename = fileURLToPath(import.meta.url ?? __filename);
 
 /**
  * @type {import("../../node_bindings/plonk_wasm")}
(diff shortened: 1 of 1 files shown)


js/node/node-backend.js
```diff
@@ -1,7 +1,8 @@
 import { isMainThread, parentPort, workerData, Worker } from 'worker_threads';
 import os from 'os';
 import wasm_ from '../../../_node_bindings/plonk_wasm.cjs';
-let filename = import.meta.url?.slice(7) ?? __filename;
+import { fileURLToPath } from 'url';
+let filename = fileURLToPath(import.meta.url);
```
mitschabaude, line 5: nice! this won't work in the commonjs version though - can you cover the case where `import.meta.url` is undefined and use `__filename` in that case? thanks

mitschabaude (commented):
thanks for contributing an important fix! just one small comment to address

js/node/node-backend.js
wizicer: Nice catch, I should considered this.

js/node/node-backend.js
```diff
@@ -1,7 +1,8 @@
 import { isMainThread, parentPort, workerData, Worker } from 'worker_threads';
 import os from 'os';
 import wasm_ from '../../../_node_bindings/plonk_wasm.cjs';
-let filename = import.meta.url?.slice(7) ?? __filename;
+import { fileURLToPath } from 'url';
+let filename = fileURLToPath(import.meta.url ?? __filename);
```
mitschabaude, line 5:
```suggestion
let url = import.meta.url;
let filename = url !== undefined ? fileURLToPath(url) : __filename;
```
unfortunately it's slightly more complicated ^^

mitschabaude (approved):

PR #22, opened 2023-05-22
Remove of field class

mitschabaude:
- delete old field class
- update bindings
- fixup
- add back the most important methods
- add back some static methods used by group
- bindings


mitschabaude:
this was premature and is superseded by https://github.com/o1-labs/snarkyjs/pull/974

PR #23, opened 2023-05-22
Improve stack traces

mitschabaude:
Small improvement to a particularly nasty type of error thrown from the Wasm bindings


kimchi/js/bindings.js
```diff
@@ -499,7 +499,11 @@ var caml_fp_vector_emplace_back = function (v, x) {
 
 // Provides: caml_fp_vector_get
 var caml_fp_vector_get = function (v, i) {
-    return new joo_global_object.Uint8Array(v[i + 1]);
+    var value = v[i + 1];
+    if (value === undefined) {
+      throw Error('caml_fp_vector_get: Index out of bounds, got ' + i + '/' + (v.length - 1));
+    }
+    return new joo_global_object.Uint8Array(value);
```
mitschabaude, line 506: this used to sometimes return `new Uint8array(undefined)` which is an empty Uint8array, which then caused an error at some later point in the code, usually "failed to fill the whole buffer"

MartinMinkov (approved):

mrmr1993 (approved):

PR #26, opened 2023-05-29
Lower requested WASM memory for iOS user agents

MartinMinkov:
# Description
:link: Related Issue: https://github.com/o1-labs/snarkyjs/issues/823

This pull request addresses an issue where SnarkyJS applications encounter an `Out of Memory` error on iOS devices, specifically iPhones and iPads, while trying to initialize the WASM memory for plonk_wasm.js. This problem seems to be unique to iOS devices as desktop Safari doesn't appear to be affected.

The solution involves modifying the maximum requested WASM memory for iOS user agents. Through testing, it's been found that these devices can handle up to 1GiB of requested memory. Attempting to request 2GB results in the same memory error. However, for all other user agents, the memory request remains unchanged, aiming for the maximum available, which is 4GiB.

### DIFF
--- a/js/web/web-backend.js
+++ b/js/web/web-backend.js
@@ -18,7 +18,9 @@ let init = wasm.default;
 let worker;
 
 async function initSnarkyJS() {
-  let { memory } = await init();
+  const memory = allocateWasmMemoryForUserAgent(window.navigator.userAgent);
+
+  await init(undefined, memory);
   let module = init.__wbindgen_wasm_module;
   let numWorkers = await getEfficientNumWorkers();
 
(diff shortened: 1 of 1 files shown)


shimkiv (approved):
This particular issue was solved 🎉 
Yet, as it was expected it revealed others, starting with inability to compile contract (which causes page reload for some reason and losing all the logs). Probably also related to memory.

js/web/web-backend.js
```diff
@@ -162,3 +164,21 @@ async function workerCall(worker, type, message) {
   worker.postMessage({ type, id, message });
   return (await promise).result;
 }
+
+function allocateWasmMemoryForUserAgent(userAgent) {
+  const isIOSDevice = /iPad|iPhone|iPod/.test(userAgent);
+  const isSafari = /Safari/.test(userAgent);
+  if (isIOSDevice && isSafari) {
```
mitschabaude, line 171: IIRC, the OOM occurred for all user agents on iOS (they all use the same Webkit engine under the hood), so should we remove the second condition?

mitschabaude (approved):
Looks good! I don't think this needed a change of the compiled bindings, btw, since it's only changing the JS side.

@MartinMinkov Yeah, a snarkyjs PR should just check out the submodule at the last commit of this PR. Ideally they are merged at the same time

Trivo25 (approved):
Nice! I think we should add this to the changelog when we get to it (dont need a seperate PR, just do it whenever)

js/web/web-backend.js
shimkiv: I think the `Safari` will be present in userAgents anyway.

PR #29, opened 2023-06-01
Prepend /bin/sh when running dump-test-vectors.sh

L-as:
While this should be redundant, it is not in my experience. Though I had the execution permission set, my system did not want to execute this script. It is not clear to me why, since the error would only say "file not found", hence I opted for this simple workaround.

### DIFF
--- a/crypto/test_vectors/dump-test-vectors.sh
+++ b/crypto/test_vectors/dump-test-vectors.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 set -e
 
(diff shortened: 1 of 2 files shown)


L-as:
I found the real issue! Let me change the PR.

L-as:
@mitschabaude This should work on all platforms that worked before, as all platforms that had /bin/bash should have /bin/sh too, but FWIW I don't see how the script could be executed on Windows given that it depends on Bash.

mitschabaude:
cool, thanks!

PR #31, opened 2023-06-02
Fix Wasm build

mitschabaude:
Fixes https://github.com/o1-labs/snarkyjs/issues/945

Makes sure that the `Cargo.lock` file gets copied to `_build` by dune, so that it is actually used for the Rust -> wasm build


kimchi/wasm/src/srs.rs
```diff
@@ -133,14 +133,13 @@ macro_rules! impl_srs {
                 let x_domain = EvaluationDomain::<$F>::new(domain_size as usize).ok_or_else(|| {
                     JsValue::from_str("caml_pasta_fp_urs_lagrange_commitment")
                 })?;
-
-                {
+                crate::rayon::run_in_pool(|| {
```
bkase, line 136: Is this important for part of this fix? If it's unrelated, can you follow up in another PR?

mitschabaude: it's related insofar as it fixes a snarkyjs test in mina CI

crypto/test_vectors/dump-test-vectors.sh
```diff
@@ -2,10 +2,11 @@
 
 set -e
 
-pushd ../../../../../crypto/proof-systems/poseidon/export_test_vectors
+DIR=$(pwd)
+cd ../../../../../crypto/proof-systems/poseidon/export_test_vectors
```
mrmr1993, line 6: Revert, use `pushd` and `popd` instead.

kimchi/wasm/src/srs.rs
mrmr1993: Can you describe what you're fixing here and why this fixes it? 

It looks like this runs in direct contradiction to the comment below. We probably need a mutex or similar on the SRS to do this safely.

crypto/test_vectors/dump-test-vectors.sh
mitschabaude: this fails because it's run with `/bin/sh` and running with `/bin/bash` causes problems for other contributors: https://github.com/o1-labs/snarkyjs-bindings/pull/29

kimchi/wasm/src/srs.rs
mitschabaude: I know, I was worrying about the comment as well.. It fixes this error when running `snarkyjs/tests/integration/inductive-proofs.js`, which failed because it tried to use rayon in this function and there was no thread pool

mitschabaude: is the function expensive? otherwise we could run it single-threaded somehow

crypto/test_vectors/dump-test-vectors.sh
mrmr1993: Because of NixOS? The usual solution for that is `#!/usr/bin/env bash`  (or just not using NixOS) instead of falling all the way back to the Bourne Shell.

kimchi/wasm/src/srs.rs
mrmr1993: > is the function expensive?

Yes.

> I know, I was worrying about the comment as well..

I suppose it's good enough for now, but we should switch to using a mutex ASAP. Not in this PR though.

crypto/test_vectors/dump-test-vectors.sh
mitschabaude: nice, I can change it to that

mrmr1993 (approved):

PR #34, opened 2023-06-05
Remove uses of JS classes in OCaml, Part 1

mitschabaude:
remove most uses of JS classes from snarkyjs-ocaml, except the field and bool classes themselves


ocaml/lib/snarky_js_bindings_lib.ml
```diff
@@ -22,7 +22,7 @@ let _console_dir s : unit =
 let raise_error s =
   Js.Js_error.(raise_ @@ of_error (new%js Js.error_constr (Js.string s)))
 
-let raise_errorf fmt = Core_kernel.ksprintf raise_error fmt
+let _raise_errorf fmt = Core_kernel.ksprintf raise_error fmt
```
Trivo25, line 25: if these arent used anymore, should we just remove them? I noticed `_console_log`also isnt used anymore

Trivo25: > well, for _console_log, I always use that if I need to debug something

Ah okay, that makes sense!

PR #35, opened 2023-06-06
Move low-level operations to JS 

Trivo25:


ocaml/lib/snarky_js_bindings_lib.ml
```diff
@@ -1227,6 +1239,8 @@ let snarky =
       object%js
         method add = Snarky.Group.add
 
+        method ecadd = Snarky.Group.ec_add
```
Trivo25, line 1242: TIL jsoo doesn't like underscores in names; naming it `ec_add` can't be resolved for some reason

mitschabaude (approved):

PR #36, opened 2023-06-07
Remove uses of JS classes in OCaml, part 2

mitschabaude:
companion of https://github.com/o1-labs/snarkyjs/pull/974


MartinMinkov (approved):
LGTM!

PR #37, opened 2023-06-07
Fixup bindings for hex serialization

mrmr1993:
This PR updates the bindings for the latest changes to the proof-systems repo.

This
* uses the `hex` option for generating the poseidon test vectors
* updates the poseidon test to work correctly with hex strings as input and output
  - previously input was base-10, but output was hex!
* updates the serde rules to use `is_human_readable: false`, so that we don't generate hex strings for field elements across the FFI boundary


querolita (approved):
Looks good to me 👍🏻

mitschabaude (commented):
the serialization changes look good, but CI now fails in `caml_pasta_fp_proof_of_rust`:
https://github.com/o1-labs/snarkyjs/actions/runs/5203561513/jobs/9386575314

nit: you don't auto-format the code with `prettier`, maybe we should add a quick command to do so. but not important

mitschabaude (dismissed):
snarkyjs CI should work before landing this

mitschabaude (approved):

PR #18, opened 2023-05-10
Wasm custom gates, revived

mitschabaude:
This restores changes to the Wasm bindings done in https://github.com/o1-labs/snarkyjs-bindings/pull/8 and reverted in https://github.com/o1-labs/snarkyjs-bindings/pull/17

mina: https://github.com/MinaProtocol/mina/pull/12524

TODO: connect this to Mina + snarkyjs branches and run CI

PR #38, opened 2023-06-07
Wasm custom gates - revived edition

rbonichon:
Revival of #8. Supersedes #18


mitschabaude:
heads up: I did merge https://github.com/o1-labs/snarkyjs-bindings/pull/38/commits/8a1fd7fbbcaad53aa5df5316bca1feb01ac2133e into this branch, after all, so that these changes are not in conflict with the reverting of the same changes that we merged into main. the unreverting merge was clean, no extra changes, just more resilient git history :D

mitschabaude (approved):

dannywillems:
Note from a call with @rbonichon: bindings.js does not include changes for the lookup verifier index.

kimchi/wasm/src/plonk_verifier_index.rs
```diff
@@ -219,12 +520,14 @@ macro_rules! impl_verification_key {
                 #[wasm_bindgen(skip)]
                 pub evals: WasmPlonkVerificationEvals,
                 pub shifts: WasmShifts,
-                // TODO: add lookup index field
+                #[wasm_bindgen(skip)]
+                pub lookup_index: Option<WasmLookupVerifier>
```
dannywillems, line 524: I think JavaScript accepts not passing the parameter to the constructor, see `caml_plonk_verifier_index_to_rust` as it will set it to undefined (?). I would add a TODO in bindings.js.

mitschabaude: yes if you pass fewer parameters in JS then `undefined` is passed instead

rbonichon: This is now fixed

rbonichon:
> Note from a call with @rbonichon: bindings.js does not include changes for the lookup verifier index.

Now it does. This PR has been re-built on #114

PR #41, opened 2023-06-13
Foreign field operations

mitschabaude:
companion to https://github.com/o1-labs/snarkyjs/pull/985


kimchi/js/bindings.js
```diff
@@ -1149,8 +1149,10 @@ var caml_plonk_gate_of_rust = function (gate) {
 // Provides: caml_fp_plonk_gate_to_rust
 // Requires: plonk_wasm, caml_plonk_wires_to_rust, caml_u8array_vector_to_rust_flat_vector
 var caml_fp_plonk_gate_to_rust = function (gate) {
+    var gateType = gate[1];
+    if (gateType > 6) gateType = gateType + 4;
     return new plonk_wasm.WasmFpGate(
-        gate[1],
+        gateType,
```
mitschabaude, line 1155: this is a temportary hack until the following proof systems fix is upstreamed to this PR: https://github.com/o1-labs/proof-systems/pull/1091

ocaml/lib/snarky_js_bindings_lib.ml
```diff
@@ -294,6 +294,35 @@ module Snarky = struct
...
+    type t_const = Impl.field FF.Element.Standard.limbs_type
+
+    (* high-level API of self-contained methods which do all necessary checks *)
+
+    let assert_valid_element (x : t) (p : t_const) : unit =
+      let external_checks = External_checks.create (module Impl) in
+      let _ = FF.valid_element (module Impl) external_checks x p in
+      FF.constrain_external_checks (module Impl) external_checks p
+
+    let add (x : t) (y : t) (p : t_const) : t =
```
querolita, line 314: Oh, so by default you include all checks of additions? This means users of the API won't be able to chain additions efficiently, and will instead require 2 rows per FFAdd instead of 1 asymptotically.

querolita (commented):
Just a question

ocaml/lib/snarky_js_bindings_lib.ml
mitschabaude: yes exactly!

querolita (approved):

kimchi/js/bindings.js
mitschabaude: reverted!

mitschabaude:
no longer needed

PR #53, opened 2023-06-21
Add new hashing functions (SHA & Keccak)

MartinMinkov:
# Description
Implements https://github.com/o1-labs/rfcs/pull/9

This PR adds support for hashing using various SHA algorithms and Keccak that is exposed in the OCaml bindings to be used by SnarkyJS. This is the layer in which the calls to Kimchi are made.

:link: SnarkyJS https://github.com/o1-labs/snarkyjs/pull/999


ocaml/lib/snarky_bindings.mli
```diff
@@ -8,6 +8,10 @@ module Poseidon : sig
   type sponge
 end
 
+(**
+This type definition is copy/pasted from the `snarkyjs_bindings.ml` using an LSP to see the type definition of `snarky`.
```
MartinMinkov, line 12: Is this the best way to describe how to add to this type? :thinking:

ocaml/lib/snarky_bindings.ml
```diff
@@ -236,6 +236,24 @@ module Poseidon = struct
...
+
+  let field_bytes_of_hex hex =
+    Array.of_list
+      (Kimchi_gadgets.Common.field_bytes_of_hex
+         (module Impl)
+         (Js.to_string hex) )
+
+  let check_bits value bits =
+    Kimchi_gadgets.Lookup.less_than_bits (module Impl) ~bits value ;
+    Kimchi_gadgets.Range_check.bits64 (module Impl) Impl.Field.zero
```
MartinMinkov, line 254: The reason for calling `Kimchi_gadgets.Range_check.bits64` is because `less_than_bits` expects a "dummy gate" to be initialized before calling it (as noted [here](https://github.com/MinaProtocol/mina/blob/bd1688a5861ff4431dbaab06e0d39129d259c307/src/lib/crypto/kimchi_backend/gadgets/lookup.ml#L31)). Calling this every time `check_bits` is called seems wasteful, is there a better way to initialize the range check once?

mitschabaude:
superseded by https://github.com/o1-labs/o1js/pull/1291

PR #58, opened 2023-06-27
Post refactor clean up: Split OCaml libs

mitschabaude:
This PR splits up the OCaml bindings into several different files:
- The main file only contains `Snarky`
- `pickles_bindings.ml` which exports `Pickles`
- `local_ledger.ml` which exports `Ledger`
- `consistency_test.ml` which exports `Test`

Also, in support of https://github.com/o1-labs/snarkyjs/pull/1004, remove unchecked Poseidon (but keep unchecked hashToGroup as a consistency test)


ocaml/lib/snarky_js_bindings_lib.ml
```diff
@@ -1,58 +1,12 @@
+open Core_kernel
```
Trivo25, line 1: Nice! Should we name this file something like `snarky_bindings.ml` then? wdyt

mitschabaude: Good idea! How do you like it now? I also removed the top level `Snarky` module in `snarky_bindings.ml` now because there's nothing else in that file

Trivo25 (approved):

PR #63, opened 2023-06-29
Always call `console_error_panic_hook::set_once` and remove ad-hoc uses of it

L-as:
The WASM module now exports `console_error_panic_hook_set_once` which is an alias for `console_error_panic_hook::set_once` to be called from the JS side. It is always called when loading the module in the wrapper JS module.


mitschabaude:
very nice! It would be good to have an accompanying snarkyjs PR so that we can see this working in CI

js/wrapper.js
```diff
@@ -5,6 +5,8 @@ export { getSnarky, getWasm, withThreadPool };
 
 let getSnarky = () => snarky;
 
+wasm.console_error_panic_hook_set_once();
```
mitschabaude, line 8: seems like a good place for this to live 👍🏻

PR #74, opened 2023-07-10
[develop] Define wasm runtime table

dannywillems:
Built on top of #73 

This PR define the structure `WasmFpRuntimeTable` and `WasmFqRuntimeTable` to be used in later PR.

PR #75, opened 2023-07-10
Define converters js wasm runtime table

dannywillems:
Built on top of #74


dannywillems:
Closing as it is merged in #79

PR #82, opened 2023-07-11
Add lookup table to prover index creation

dannywillems:
Built on top of #79


dannywillems:
Oops merged. Reopening...

PR #83, opened 2023-07-12
Instruction npm clean

dannywillems:


README.md
```diff
@@ -51,3 +51,6 @@ This command will
 - update the `MINA_COMMIT` file
 
 > ℹ️ Use `npm run bindings` to update the source tree after you have finished a batch of work / a PR which touches the OCaml/Rust layers. SnarkyJS CI depends on these artifacts, so this is a required step before merging such changes.
+
+After submitting a patch updating the OCaml/Rust bindings, you might need to run
+`npm run clean` in `snarkyjs` before running the unit tests locally.
```
mitschabaude, line 56: I feel it would be more accurate to put that under the explanation of `npm run make`. That's because the `make` script doesn't clean `dist`, while the `bindings` script does

mitschabaude:
Or actually, maybe the better fix is to make the `npm run make` script cleanup `dist`?

dannywillems:
> Or actually, maybe the better fix is to make the `npm run make` script cleanup `dist`?

Done.

PR #85, opened 2023-07-13
[develop] Bindings - pass runtime tables to proof creation

dannywillems:
See explanation in https://github.com/MinaProtocol/mina/pull/13593


kimchi/js/bindings.js
```diff
@@ -2279,11 +2279,28 @@ var caml_pasta_fp_proof_of_rust = function (x) {
   return [0, messages, proof, evals, ft_eval1, public_, prev_challenges];
 };
 
+// Provides: caml_fp_runtime_table_to_rust
+// Requires: plonk_wasm, caml_fp_vector_to_rust
+var caml_fp_runtime_table_to_rust = function (caml_runtime_table, mk_class) {
+  // A value caml_runtime_table is a record on the OCaml side.
+  // The converter should be changed if CamlRuntimeTable is modified.
```
mrmr1993, line 2286: I think this is true of all bindings in all cases. Delete?

mrmr1993 (approved):

kimchi/js/bindings.js
dannywillems: No strong opinion. True it is not that useful, removed in b3382c43c4df992436e843fe33c3d9559c826da0. I think keeping a small comment saying what the variable contains is nice to have, to avoid going back to the Rust code.
I'll remove the others later.

dannywillems:
Squashing fixup commits and regenerating bindings as mina changed too and has been rebased.

PR #90, opened 2023-07-23
[develop] Update bindings for optional column evaluations

mrmr1993:
This is the counterpart to https://github.com/MinaProtocol/mina/pull/13685.


mitschabaude (changes requested):
we should merge this in lockstep with the snarkyjs part

mitschabaude (approved):

PR #91, opened 2023-07-23
[berkeley] Expose and use overrideWrapDomain

mrmr1993:
This PR is the counterpart to https://github.com/MinaProtocol/mina/pull/13546.

PR #92, opened 2023-07-23
[develop] Expose and use overrideWrapDomain

mrmr1993:
This PR is #91 for develop. This is the counterpart to https://github.com/MinaProtocol/mina/pull/13686.


ocaml/lib/pickles_bindings.ml
```diff
@@ -461,13 +463,26 @@ let pickles_compile (choices : pickles_rule_js array)
   (* translate method circuits from JS *)
   let public_input_size = signature##.publicInputSize in
   let public_output_size = signature##.publicOutputSize in
+  let override_wrap_domain =
+    match Js.Optdef.to_option signature##.overrideWrapDomain with
```
dannywillems, line 467: Berkeley uses
```
  let override_wrap_domain =
    Js.Optdef.to_option signature##.overrideWrapDomain
    |> Option.map ~f:Pickles_base.Proofs_verified.of_int
  in
```

mitschabaude: updated!

dannywillems: I make a commit changing to it, feel free to remove it if you do not agree with

mitschabaude (approved):

ocaml/lib/pickles_bindings.ml
dannywillems: Too late :grimacing:. You're a more morning person than me :)

dannywillems (approved):

ocaml/lib/pickles_bindings.ml
mitschabaude: I already made that commit :D

PR #93, opened 2023-07-23
[berkeley] Remove special-casing of lookup-related columns

mrmr1993:
This PR updates the bindings for https://github.com/MinaProtocol/mina/pull/13547.


kimchi/wasm/src/plonk_proof.rs
```diff
@@ -699,7 +699,10 @@ macro_rules! impl_proof {
                     coefficients: array_init(|_| eval()),
                     z: eval(),
                     s: array_init(|_| eval()),
-                    lookup: None,
+                    lookup_aggregation: None,
+                    lookup_sorted: [None, None, None, None, None],
```
dannywillems, line 703: The order is not the same than in [proof-systems](https://github.com/o1-labs/proof-systems/blob/21ae80d5f624480b68eca75f08c558ff603edcd0/book/src/specs/kimchi.md), but I think it should be fine:

dannywillems (approved):

PR #108, opened 2023-08-02
Plonk_proof: do not use std::array but use array_init

dannywillems:
To avoid introducing a new import. `array_init` is used in the rest of the file.

### DIFF
--- a/kimchi/wasm/src/plonk_proof.rs
+++ b/kimchi/wasm/src/plonk_proof.rs
@@ -16,7 +16,6 @@ use ark_ff::One;
 use array_init::array_init;
 use kimchi::circuits::wires::COLUMNS;
 use kimchi::verifier::Context;
-use std::array;
 // use std::path::Path;
 use groupmap::GroupMap;
 use kimchi::proof::{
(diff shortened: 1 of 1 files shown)


kimchi/wasm/src/plonk_proof.rs
```diff
@@ -690,7 +689,7 @@ macro_rules! impl_proof {
                     rot_selector: None,
                     lookup_aggregation: None,
                     lookup_table: None,
-                    lookup_sorted: array::from_fn(|_| None),
+                    lookup_sorted: array_init(|_| None),
```
rbonichon, line 692: Why? Is it somehow more efficient to use `array_init`? If not, I think it's better to used the standard library.

rbonichon: I would tend to exactly do the contrary, i.e., change all calls to `array_init` to use `from_fn` to remove the dependency on a given `crate`

dannywillems: Mmh, I don't have any real reason except than staying consistent. I assumed there was a good reason before to use `array_init` everywhere.

PR #116, opened 2023-08-16
Reduce loading time

mitschabaude:
supports https://github.com/o1-labs/snarkyjs/pull/1073


crypto/bindings.ts
```diff
@@ -0,0 +1,9 @@
+import { prefixHashes, prefixHashesLegacy } from '../crypto/constants.js';
+
+const tsBindings = {
+  prefixHashes,
+  prefixHashesLegacy,
+};
+
+// this is put in a global variable so that ../kimchi/js/bindings.js finds it
+(globalThis as any).__snarkyTsBindings = tsBindings;
```
mitschabaude, line 9: this is the new TS bindings file, [explained here](https://github.com/o1-labs/snarkyjs/pull/1073#discussion_r1300137784)

crypto/constants.ts
```diff
@@ -14,6 +14,255 @@ let prefixes = {
   "zkappUri": "MinaZkappUri********",
   "deriveTokenId": "MinaDeriveTokenId***"
 };
+let prefixHashes = {
+  "MinaReceiptUC*******": [
+    "17890046087060600378523962727948914890955845280331606421027576706277370812052",
+    "2013454333407534591106411126192774509000627947643864003868625362325108814274",
+    "11843535353803037930790320689348405921362689396018494249827525941843824404081"
+  ],
```
mitschabaude, line 22: hard-coded hashes exposed to jsoo via the new TS bindings. we hard-code them to avoid computing them at the top level

kimchi/js/bindings.js
```diff
@@ -4,6 +4,15 @@
...
 */
 
+// Provides: tsBindings
+var tsBindings = globalThis.__snarkyTsBindings;
+
+// Provides: getTsBindings
+// Requires: tsBindings
+function getTsBindings() {
+    return tsBindings;
+}
```
mitschabaude, line 14: expose new bindings file to jsoo

ocaml/snarky_js_constants.ml
```diff
@@ -18,6 +23,79 @@ let prefixes =
...
+        (s, Random_oracle.Legacy.(State.to_array (salt s)))
+  in
+  ((s :> string), array field fields)
+
+let prefix_hashes =
+  let open Hash_prefixes in
+  `Assoc
+    (List.map ~f:(prefix_hash_entry Kimchi)
+       [ (receipt_chain_user_command :> string)
+       ; (receipt_chain_zkapp :> string)
```
mitschabaude, line 43: these changes are about computing some hashes at compile time and put them in the `constants.ts` file

js/web/web-backend.js
```diff
@@ -11,22 +11,23 @@ import snarkyJsWebSrc from 'string:../../../web_bindings/snarky_js_web.bc.js';
...
  */
-let worker;
+let numWorkers = undefined;
 
 async function initSnarkyJS() {
   const memory = allocateWasmMemoryForUserAgent(navigator.userAgent);
   await init(undefined, memory);
 
   let module = init.__wbindgen_wasm_module;
-  let numWorkers = await getEfficientNumWorkers();
```
mitschabaude, line 25: change 1 to this file: we don't call `getEfficientNumWorkers()` right away, because it's very expensive (inspects the device GPU via WebGL). Instead, we call it before we need workers the first time, which is likely a slow operation anyway

js/web/web-backend.js
```diff
@@ -11,22 +11,23 @@ import snarkyJsWebSrc from 'string:../../../web_bindings/snarky_js_web.bc.js';
...
-
-  worker = inlineWorker(srcFromFunctionModule(mainWorker));
-  await workerCall(worker, 'start', { memory, module, numWorkers });
-  globalThis.plonk_wasm = overrideBindings(wasm, worker);
```
mitschabaude, line 29: change 2 to this file: creating the main worker and waiting for its promise to return (which is after all its child workers are started) is very slow. Therefore, we do this in a way that is not blocking our top-level await. This is fine, the main worker will only be needed when we do our first expensive operation and we await the promise at that time

ocaml/lib/consistency_test.ml
```diff
@@ -84,30 +84,33 @@ end
 (* deriver *)
 let account_update_of_json, _account_update_to_json =
   let deriver =
-    Account_update.Graphql_repr.deriver @@ Fields_derivers_zkapps.Derivers.o ()
+    lazy
+      ( Account_update.Graphql_repr.deriver
+      @@ Fields_derivers_zkapps.Derivers.o () )
```
mitschabaude, line 89: we compute a bunch of stuff lazily in `consistency_test.ml` and `local_ledger.ml` which was computed at the top level before

ocaml/lib/snarky_bindings.ml
```diff
@@ -203,13 +203,9 @@ module Poseidon = struct
   module Poseidon_sponge =
     Sponge.Make_sponge (Sponge.Poseidon (Pickles.Tick_field_sponge.Inputs))
 
-  let sponge_params_checked =
-    Sponge.Params.(
-      map pasta_p_kimchi
-        ~f:(Fn.compose Impl.Field.constant Impl.Field.Constant.of_string))
+  let sponge_params = Kimchi_pasta_basic.poseidon_params_fp
```
mitschabaude, line 206: don't recompute stuff that is computed in elsewhere already

mitschabaude (commented):
some explanations for reviewers!

js/web/web-backend.js
MartinMinkov: Great optimization! Great for cases where we only need constants since that can all be done in JS, right?

crypto/constants.ts
MartinMinkov: Nit: If we could throw some inline comments when this (or similar) file(s) is generated that point to the OCaml file, it would really help context for other developers if debugging or adding features in this area.

crypto/bindings.ts
MartinMinkov: Nit: Depending how you feel on comments, you could argue the explanation could be added here, mostly just elaborating on the inline comment you have already. Could be helpful for others look at this file and wondering more, wdyt?

MartinMinkov (approved):
Great use of lazy and compiling prefixes ahead of time instead of hitting runtime cost! 

I mostly just had comments about documentation, feel free to address or make issues for later

ocaml/lib/dune
```diff
@@ -58,6 +58,7 @@
   ;; js-specific overrides ;;
   cache_dir.fake
   digestif.ocaml
+  hash_prefix_create.js
```
Trivo25, line 61: oh you can just import `js` file like that :o

Trivo25 (approved):

js/web/web-backend.js
mitschabaude: Yes!

ocaml/lib/dune
Trivo25: oh i was confused for a second! lol

PR #121, opened 2023-08-21
[berkeley] Add lookup verifier index

dannywillems:


kimchi/js/bindings.js
```diff
@@ -1565,33 +1565,26 @@ var caml_plonk_verification_evals_to_rust = function (
   x,
   klass,
   poly_comm_class,
-  mk_affine
+  mk_affine,
```
rbonichon, line 1568: I think we don't need either of mk_affine and `poly_comm_class` once we have `poly_comm_to_rust`

rbonichon (approved):
Only 1 minor thing to change. LGTM

kimchi/js/bindings.js
dannywillems: Good catch! Fixed.

PR #123, opened 2023-08-22
Ignore warning about unused value

dannywillems:
And run `dune build @fmt`


dannywillems:
Must be removed later, see https://github.com/MinaProtocol/mina/issues/13914.

PR #127, opened 2023-08-24
Reimplement snarky backend with TS bigints

mitschabaude:
We swap out the part of the snarky backend which doesn't inherently need kimchi with TS implementations. this brings a massive speed-up to non-kimchi operations such as witness generation.

The main complexity introduced by this change is the need to convert between bigints and Rust values in methods that still use kimchi. Since the new backend is implemented in modern TS (not JSOO-friendly JS), I chose to move Rust/JSOO conversion functions to TS as well. This should make it much easier to work with these bindings in the future, ~~but comes at the price of a quite massive PR~~. EDIT: Move of conversion to TS was split into several incremental PRs which this one sits on top of

TODOs:
- [x] make ts bindings integration work with jest, to fix unit tests in https://github.com/o1-labs/snarkyjs/pull/1092
- [x] add consistency tests between new backend and kimchi backend
- [x] add benchmarks for more complex scenarios (like proving)
- [x] (optional) split bindings.js and conversion functions into more files


Trivo25 (approved):
First pass looks good! I am pre-approving so you dont have to wait for me. I'll do another pass later tonight

crypto/bindings/field.ts
```diff
@@ -95,12 +95,13 @@ function createFieldBindings(Field: FiniteField) {
     random(): Field {
       return [0, Field.random()];
     },
+    // TODO
```
dannywillems, line 98: ?

crypto/bindings/field.ts
```diff
@@ -95,12 +95,13 @@ function createFieldBindings(Field: FiniteField) {
...
     rng(i: number): Field {
       console.warn('rng is not implemented');
       return [0, Field.random()];
     },
-    to_bigint(x: Field): Bigint256 {
```
dannywillems, line 102: ?

crypto/bindings/vector.ts
```diff
@@ -27,7 +27,7 @@ const FieldVectorBindings = {
         `FieldVector.get(): Index out of bounds, got ${i}/${v.length - 1}`
       );
     }
-    return value;
+    return [...value];
```
dannywillems, line 30: ?

crypto/bindings/field.ts
dannywillems: Ok, I see. Can you add a comment in the code explaining this please? For our future us :).

crypto/bindings/vector.ts
dannywillems: So does it make a deep copy?

crypto/bindings/field.ts
dannywillems: Don't we want a deep copy? I.e. copying x deeply too?

crypto/bindings/field.ts
dannywillems: Can you add an explanation next to the TODO and link with an issue? It will be better to track it then.

crypto/bindings/vector.ts
mitschabaude: yeah exactly

crypto/bindings/field.ts
mitschabaude: x is a `bigint` which is a JS primitive value. it can't be mutated, so no need (and no way) to deep copy

crypto/bindings/field.ts
mitschabaude: ok! I think I'll just remove the TODO and make the function throw an error. always better to avoid doing work that has no impact

dannywillems (approved):

PR #128, opened 2023-08-25
Custom gates support 

rbonichon:
Bindings for https://github.com/MinaProtocol/mina/pull/14021


dannywillems:
The change in bindings_js_test.ml has been added to develop with https://github.com/MinaProtocol/mina/pull/13991.
The changes that still needs to be added are the commitments.

mrmr1993 (approved):

kimchi/js/bindings.js
```diff
@@ -1714,6 +1732,12 @@ var caml_plonk_verification_evals_of_rust = function (x, affine_klass) {
     mul_comm,
     emul_comm,
     endomul_scalar_comm,
+    xor_comm,
+    range_check0_comm,
+    range_check1_comm,
+    foreign_field_add_comm,
+    foreign_field_mul_comm,
+    rot_comm,
     0,
```
mitschabaude, line 1741: the 0 at the end here seems not to match the `kimchi_types.ml` [type](https://github.com/MinaProtocol/mina/pull/13858/files#diff-a1db9c25d545f2f561ae6daff600b8dd51c67e8faaefcc4b0858ff098c8b2f82R235-R240), maybe a relic of some older version where `None` was passed here?

mitschabaude (approved):
approving with one outstanding comment!

kimchi/js/bindings.js
rbonichon: 🤷 I don't know. It was there already some I kept it - and it does not seem like it's going to trigger any bug. But looking twice, it does not make much sense indeed. Let me update the code.

PR #136, opened 2023-09-07
fix links

ghost-not-in-the-shell:


mitschabaude (approved):
thanks, and sorry for not fixing this when I moved the files 😅

PR #140, opened 2023-09-12
Bigint backend pt 3: Move ec point conversion to TS

mitschabaude:
Incremental step on top of #139

* [x] **Conversion of curve points and polynomial commitments**
* [ ] Conversion of verification key
* [ ] Conversion of oracles
* [ ] Conversion of proof
* [ ] Cleanup conversion modules


crypto/bindings/conversion.ts
```diff
@@ -10,9 +10,12 @@ import type {
 import type * as wasmNamespace from '../../compiled/node_bindings/plonk_wasm.cjs';
 import { bigIntToBytes, bytesToBigInt } from '../bigint-helpers.js';
 
+export { createRustConversion };
+
 type Field = Uint8Array;
 
-export { createRustConversion };
+// Kimchi_types.or_infinity
+type OrInfinity = MlOption<MlTuple<Field, Field>>;
```
dannywillems, line 18: Nit: I would add an alias saying it is a point in affine form.

mitschabaude: Isn't it more helpful to be named like the OCaml type? I think that makes it easier to compare the OCaml and TS interfaces

dannywillems: Ah! Yes, good point. It is copying `Kimchi_types.or_infinity`. It is better by being explicit then. However, to be fully correct, even though it is easy to make the conversion, it is not strictly equivalent as we have:
```
type nonrec 'f or_infinity = Infinity | Finite of ('f * 'f)
```
and not an option.

crypto/bindings/conversion.ts
```diff
@@ -47,18 +64,59 @@ function createRustConversion(wasm: wasm) {
...
+        return polyCommFromRust(polyComm, CommitmentCurve, false);
+      },
     };
   }
 
-  const fpConversion = perField(wasm.WasmFpGate);
-  const fqConversion = perField(wasm.WasmFqGate);
+  // TODO: we have to lie about types here:
+  // -) the WasmGVesta class doesn't declare __wrap() but our code assumes it
+  // -) WasmGVesta doesn't declare the `ptr` property but our code assumes it
```
dannywillems, line 95: What are the consequences of it?

crypto/bindings/conversion.ts
mitschabaude: oh that's true! the jsoo representation is the same -- but I can rewrite the type to be more clear

crypto/bindings/conversion.ts
```diff
@@ -47,18 +64,59 @@ function createRustConversion(wasm: wasm) {
...
+      polyCommFromRust(polyComm: WasmPolyComm): PolyComm {
+        return polyCommFromRust(polyComm, CommitmentCurve, false);
+      },
     };
   }
 
-  const fpConversion = perField(wasm.WasmFpGate);
-  const fqConversion = perField(wasm.WasmFqGate);
+  // TODO: we have to lie about types here:
+  // -) the WasmGVesta class doesn't declare __wrap() but our code assumes it
```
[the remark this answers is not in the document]
mitschabaude: we need to use this so that we can instantiate these classes from pointers. It's sad that the autogenerated Wasm types don't reflect the full functionality, but there are no bad consequences other than having to coerce the type rather than use the declared type, in this one line

crypto/bindings/conversion.ts
rbonichon: The runtime representation is indeed the same :-)

crypto/bindings/conversion.ts
mitschabaude: same as above!

crypto/bindings/conversion.ts
```diff
@@ -98,3 +158,120 @@ function fieldsFromRustFlat(fieldBytes: Uint8Array): MlArray<Field> {
...
+
+type WasmAffine = wasmNamespace.WasmGVesta | wasmNamespace.WasmGPallas;
+type MakeAffine<A extends WasmAffine> = () => A & { ptr: number };
+
+function affineFromRust(pt: WasmAffine): OrInfinity {
+  if (pt.infinity) {
+    pt.free();
+    return 0;
+  } else {
+    let x = fieldFromRust(pt.x);
```
dannywillems, line 172: Might be obvious, but still asking.
```
function fieldFromRust(x: Uint8Array): Field {
  return x;
}
```
This does create a branch new value right? i.e. the free on the point won't free the field element?

dannywillems: So it is going to access a freed value atm, and you plan to update it later?

kimchi/js/bindings.js
```diff
@@ -2363,24 +2226,24 @@ var caml_pasta_fp_opening_proof_of_rust = function (x) {
...
 // Provides: caml_fp_lookup_commitments_to_rust
-// Requires: plonk_wasm, caml_vesta_poly_comm_to_rust, js_class_vector_to_rust_vector, caml_opt_to_rust
+// Requires: plonk_wasm, tsRustConversion, js_class_vector_to_rust_vector, caml_opt_to_rust
 var caml_fp_lookup_commitments_to_rust = function (caml_lc) {
   var convertArray = function (v) {
     var n = v.length - 1;
     var res = new Array(n);
     for (var i = 0; i < n; ++i) {
-      res[i] = caml_vesta_poly_comm_to_rust(v[i + 1]);
+      res[i] = tsRustConversion.fp.polyCommFromRust(v[i + 1]);
```
dannywillems, line 2235: ToRust?

kimchi/js/bindings.js
```diff
@@ -2808,7 +2670,8 @@ var caml_pasta_fq_proof_of_rust = function (x) {
     res[1] = tsRustConversion.fq.vectorFromRust(
       prev_challenges_scalars.get(i - 1)
     );
-    res[2] = caml_pallas_poly_comm_of_rust(prev_challenges_comms[i]);
+    // FIXME (gregor): this accesses prev_challenges_comms at one index too high
+    res[2] = tsRustConversion.fq.polyCommFromRust(prev_challenges_comms[i]);
```
dannywillems, line 2674: Ah? Already before?

crypto/bindings/conversion.ts
dannywillems: Yes, the runtime is the same. I would be in favour of the change for consistency and to stay in line with the native code, if it is fine for you.

crypto/bindings/conversion.ts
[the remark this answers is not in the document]
dannywillems: Does it come from `wasm-bindgen` not generating `__wrap()` in ts types? :thinking:

dannywillems:
Discussed offline with Gregor: I would be in favor or always having the corresponding o1js PR to have the tests and be sure the code is correct. Atm, we only use a linter in the CI. As the code is moved to pure TS, tests could be added. That would be important IMO.
I reviewed by looking at the correspondances between the JS and TS code.
Gregor told me the final PR tests everything. I can accept the PR as it is, modulo my comments above.

dannywillems:
> Discussed offline with Gregor: I would be in favor or always having the corresponding o1js PR to have the tests and be sure the code is correct. Atm, we only use a linter in the CI. As the code is moved to pure TS, tests could be added. That would be important IMO. I reviewed by looking at the correspondances between the JS and TS code. Gregor told me the final PR tests everything. I can accept the PR as it is, modulo my comments above.

I would open a follow issue for the tests in TS

dannywillems:
Is it possible to add in the CI the code typechecks?

crypto/bindings/conversion.ts
```diff
@@ -10,9 +10,12 @@ import type {
 import type * as wasmNamespace from '../../compiled/node_bindings/plonk_wasm.cjs';
 import { bigIntToBytes, bytesToBigInt } from '../bigint-helpers.js';
 
+export { createRustConversion };
```
dannywillems, line 13: Nit: that should not be a change (moved up, see removed line before).

crypto/bindings/conversion.ts
```diff
@@ -98,3 +158,120 @@ function fieldsFromRustFlat(fieldBytes: Uint8Array): MlArray<Field> {
...
+    return [0, [0, x, y]];
+  }
+}
+
+function affineToRust<A extends WasmAffine>(
+  pt: OrInfinity,
+  klass: MakeAffine<A>
+) {
+  var res = klass();
+  if (pt === 0) {
```
dannywillems, line 184: If you change `OrInfinity` as discussed [here](https://github.com/o1-labs/o1js-bindings/pull/140/files#r1329991174), it will still work as by luck `Infinity` is the first element of the ADT, but I would use a helper instead.

crypto/bindings/conversion.ts
```diff
@@ -98,3 +158,120 @@ function fieldsFromRustFlat(fieldBytes: Uint8Array): MlArray<Field> {
...
+    res.infinity = true;
+  } else {
+    res.x = fieldToRust(pt[1][1]);
```
dannywillems, line 187: very small option: there will be two indirections to access the array `pt[1]` is repeated. Maybe just keep it once?

crypto/bindings/conversion.ts
```diff
@@ -98,3 +158,120 @@ function fieldsFromRustFlat(fieldBytes: Uint8Array): MlArray<Field> {
...
+type WasmPolyCommClass = wasm['WasmFpPolyComm'] | wasm['WasmFqPolyComm'];
+
+function polyCommFromRust(
+  polyComm: WasmPolyComm,
+  klass: WrapperClass<WasmAffine>,
+  shouldFree: boolean
+): PolyComm {
+  let rustShifted = polyComm.shifted;
+  let rustUnshifted = polyComm.unshifted;
+  let mlShifted: MlOption<OrInfinity> =
```
dannywillems, line 205: This would require a change if it is not an option as discussed [here](https://github.com/o1-labs/o1js-bindings/pull/140#discussion_r1329991174).
I would use helpers too.

crypto/bindings/conversion.ts
```diff
@@ -98,3 +158,120 @@ function fieldsFromRustFlat(fieldBytes: Uint8Array): MlArray<Field> {
...
+    rustShifted === undefined ? 0 : [0, affineFromRust(rustShifted)];
```
dannywillems, line 206: Can we use helpers? None values for Rust options are translated to `undefined`, and I would be in favor in making it clear in the code for our future us :).

crypto/bindings/conversion.ts
```diff
@@ -98,3 +158,120 @@ function fieldsFromRustFlat(fieldBytes: Uint8Array): MlArray<Field> {
...
+  return [0, mlUnshifted, mlShifted];
+}
+
+function polyCommToRust(
+  [, camlUnshifted, camlShifted]: PolyComm,
+  PolyComm: WasmPolyCommClass,
+  makeAffine: MakeAffine<WasmAffine>
+): WasmPolyComm {
+  let rustShifted =
+    camlShifted === 0 ? undefined : affineToRust(camlShifted[1], makeAffine);
```
dannywillems, line 222: As discussed [here](https://github.com/o1-labs/o1js-bindings/pull/140#discussion_r1329991174), would require a change, even though OK as the first element of the ADT is 0.

crypto/bindings/conversion.ts
```diff
@@ -98,3 +158,120 @@ function fieldsFromRustFlat(fieldBytes: Uint8Array): MlArray<Field> {
...
+  shouldFree: boolean
+): MlArray<TMl> {
+  var n = rustVector.length;
+  var array: TMl[] = new Array(n);
+  for (let i = 0; i < n; i++) {
+    var rustValue = klass.__wrap(rustVector[i]);
+    array[i] = convert(rustValue);
+    if (shouldFree) rustValue.free();
+  }
+  return [0, ...array];
```
dannywillems, line 256: Is it copying a new array or appending at the beginning? An optim would be to avoid this duplicate allocation and instantiate directly `Array(n + 1)`.

crypto/bindings/conversion.ts
```diff
@@ -98,3 +158,120 @@ function fieldsFromRustFlat(fieldBytes: Uint8Array): MlArray<Field> {
...
+}
+
+// TODO get rid of excessive indirection here
+
+function mlArrayToRustVector<TRust extends WrappedPointer, TMl>(
+  [, ...array]: MlArray<TMl>,
```
dannywillems, line 262: Do you know how it is translated at runtime? The first element is dropped? What is pushed on the stack when calling this function? `ptr` or `ptr + single_element_size`?

crypto/bindings/conversion.ts
```diff
@@ -98,3 +158,120 @@ function fieldsFromRustFlat(fieldBytes: Uint8Array): MlArray<Field> {
...
+  convert: (c: TMl, makeNew: () => TRust) => TRust,
+  makeNew: () => TRust
+): Uint32Array {
+  let n = array.length;
+  let rustVector = new Uint32Array(n);
+  for (var i = 0, l = array.length; i < l; i++) {
```
dannywillems, line 268:
```suggestion
  for (var i = 0; i < n; i++) {
```
?

crypto/bindings/conversion.ts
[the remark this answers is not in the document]
mitschabaude: Yes

crypto/bindings/conversion.ts
mitschabaude: This function does nothing but later it will do a conversion because on one side the field type will change.
I wouldn't say it accesses a freed value (whatever that might mean for a function that does nothing), because the field value here is already copied out of the wasm memory and is a self-contained array of bytes managed by js

mitschabaude: So what's creating the new value here is the property access, `pt.x`!

kimchi/js/bindings.js
mitschabaude: good catch!

kimchi/js/bindings.js
mitschabaude: I think so, but no need to figure this out here IMO since one of the next commits moves it to TS anyway and it's much easier to reason about these things in the new code (where the `[0, ..array]` structure is encoded in types)

crypto/bindings/conversion.ts
mitschabaude: I think it's fine, I've been adding more types and having the exports at the top is better than burying it somewhere below in the file

crypto/bindings/conversion.ts
mitschabaude: it doesn't work by luck - it works because this is all strictly typed and the type of `pt` already correctly reflected the runtime value. TS wouldn't have let us access `pt[1]` below if we didn't handle the `pt === 0` case here.

I improved readability of this code by adding a constant `Infinity = 0`. Other than that, I think handling the different possibilities of an enum with if statements, like this function does, is the correct TS analogue of matching an enum in OCaml. It's type-safe and it is idiomatic JS code. I don't want to introduce a separate helper function for every enum just to express a match statement.

crypto/bindings/conversion.ts
mitschabaude: yeah I changed it to destructure everything right away, that's more readable and doesn't repeat the array access

crypto/bindings/conversion.ts
mitschabaude: yeah will use a helper for options

crypto/bindings/conversion.ts
mitschabaude: no -- camlShifted is actually supposed to be an `Option`. I don't want to use a helper here because that would require me to create an inline conversion function in this case, which is inefficient, and `polyCommToRust` is used a lot. So will stick to the existing code

crypto/bindings/conversion.ts
mitschabaude: Yes we could slightly optimize the code, but it comes at the cost that we have to always deal with the index translation by 1. This has led to bugs in the previous code base, so I'd like to keep it simple for now and always to the conversion to/from an ml array as a separate step.
In theory we could use `array.unshift(0)` at the end to get the same result without copying the array. The only reason I don't do that is because TS doesn't understand it

crypto/bindings/conversion.ts
mitschabaude: you call the function with the full array. this statement here happens inside the function. it destructures the full array to drop the first element. so we create a copy which has 1 element less.

again, I want to avoid the error-prone inline index-shifting

mitschabaude:
> Discussed offline with Gregor: I would be in favor or always having the corresponding o1js PR to have the tests and be sure the code is correct

Running CI with o1js main + the latest version of this PR here: https://github.com/o1-labs/o1js/pull/1117

mitschabaude:
> Is it possible to add in the CI the code typechecks?

that would be very nice indeed, will think about it. probably just means an extra tsconfig file here and make CI install TS

crypto/bindings/conversion.ts
dannywillems: Interesting that it is a two steps. Not very good in terms of runtime cost to destructure then. I didn't know.
IMHO, we should avoid it as often as we can to optimise the code. We could have sometimes large arrays and copies are not as free as we can think. And it we can avoid it... why not :).

crypto/bindings/conversion.ts
dannywillems: Could we keep track in an issue of this kind of optimisations we could do? Just for the future us :).

dannywillems: Opened follow-up issue: https://github.com/o1-labs/o1js-bindings/issues/154

crypto/bindings/conversion.ts
dannywillems: I opened https://github.com/o1-labs/o1js-bindings/issues/155.

PR #142, opened 2023-09-12
Bigint backend pt 5: Move oracle conversion to TS

mitschabaude:
Incremental step on top of #141

This PR part of a multi-PR series to move JSOO / Wasm conversion methods from `bindings.js` to TS:

* [x] Conversion of wires, gates, and different kinds of field vector
* [x] Conversion of curve points and polynomial commitments
* [x] Conversion of verification key
* [x] **Conversion of oracles**
* [ ] Conversion of proof
* [ ] Cleanup conversion modules

PR #143, opened 2023-09-12
Bigint backend pt 6: Move proof conversion to TS

mitschabaude:
Incremental step on top of #142

* [x] Conversion of curve points and polynomial commitments
* [x] Conversion of verification key
* [x] Conversion of oracles
* [x] **Conversion of proof**
* [ ] Cleanup conversion modules


crypto/bindings/util.ts
```diff
@@ -21,3 +21,16 @@ function mapTuple<T extends Tuple<any>, B>(
...
+/**
+ * tuple type that has the length as generic parameter
+ */
+type MlTupleN<T, N extends number> = N extends N
+  ? number extends N
+    ? [0, ...T[]] // N is not typed as a constant => fall back to array
+    : [0, ...TupleRec<T, N, []>]
+  : never;
+
+type TupleRec<T, N extends number, R extends unknown[]> = R['length'] extends N
```
dannywillems, line 34: Oh :exploding_head:...

crypto/bindings/util.ts
```diff
@@ -21,3 +21,16 @@ function mapTuple<T extends Tuple<any>, B>(
 ): { [i in keyof T]: B } {
   return tuple.map(f) as any;
 }
+
+/**
+ * tuple type that has the length as generic parameter
+ */
+type MlTupleN<T, N extends number> = N extends N
```
dannywillems, line 28: :exploding_head:

dannywillems: I don't understand this. What is supposed to represent?

crypto/bindings/util.ts
dannywillems: Same than for MlTupleN

crypto/bindings/conversion.ts
```diff
@@ -526,6 +653,80 @@ function createRustConversion(wasm: wasm) {
       return mlRo;
     }
 
+    function commitmentsToRust(
+      commitments: ProverCommitments
```
dannywillems, line 657: IMO, it would be good to prefix with `Ml` and `Wasm` all the types to ease the reading of the code.

crypto/bindings/conversion.ts
```diff
@@ -526,6 +653,80 @@ function createRustConversion(wasm: wasm) {
...
+        self.pointsToRust(l),
+        self.pointsToRust(r),
+        self.pointToRust(delta),
+        fieldToRust(z1),
+        fieldToRust(z2),
+        self.pointToRust(sg)
+      );
+    }
+    function openingProofFromRust(proof: WasmOpeningProof): OpeningProof {
+      let [, ...l] = self.pointsFromRust(proof.lr_0);
```
dannywillems, line 716: IIUC, it ignores the first element of the list, i.e. the `0` for the Caml array memory repr?

dannywillems (approved):
LGTM. Let's add more things in the CI later please :), like typechecking, unit tests, etc :).

crypto/bindings/conversion.ts
mitschabaude: yes!

crypto/bindings/util.ts
mitschabaude: let's say you want a type for a static size tuple like `field * field * field * field * field`.
The TS version would be `[0, Field, Field, Field, Field, Field]`

this generic type let's you instead write `MlTuple<Field, 5>` so it saves you some ugliness (especially in the case of size-15 tuples that we have in some places)

mitschabaude: also let's you have functions like `mapTuple` that are generic over the tuple length but still preserve its typed size

PR #144, opened 2023-09-12
Bigint backend pt 7: Cleanup TS conversion

mitschabaude:
Incremental step on top of #143

* [x] Conversion of curve points and polynomial commitments
* [x] Conversion of verification key
* [x] Conversion of oracles
* [x] Conversion of proof
* [X] **Cleanup conversion modules**

Note for reviewers: This has a big diff, but it doesn't contain any actual code changes. It just splits up the huge `conversion.ts` file into multiple smaller files


dannywillems (commented):
Would be nice to have some headers/doc in a README to understand the split in the different conversion files. I suppose the PR is only about splitting conversion.ts.

mitschabaude:
Yes this is just about splitting up the file. It was becoming too big and hard to navigate. I hoped the PR description was enough :D

dannywillems (approved):

PR #145, opened 2023-09-13
fix Struct.toFields()

Comdex:
Remove unnecessary sorting so that the output of `Struct.toFields()` is always consistent with the output of `CircuitValue.toFields()` (So fields are output in the order in which their corresponding properties are defined).


mitschabaude:
Nice, appreciate this! I only learned that JS object property order is deterministic after I wrote this code 😅 

@Comdex since this could easily be a breaking change (order of fields affects constraints), we also need an o1js PR for this, to see CI running

Comdex:
Sorry for replying so late, I have already submitted an o1js PR https://github.com/o1-labs/o1js/pull/1124

mitschabaude:
@Comdex can you also invite me to this repo?

Comdex:
> @Comdex can you also invite me to this repo?

Done

PR #146, opened 2023-09-13
Fix lookup bindings

mitschabaude:
A couple of fixes to bindings.js. In a test that uses foreign field gates in plain Kimchi circuits in JS, these fixes are what makes compile, prove and verify run; but verify at this moment returns false


kimchi/js/bindings.js
```diff
@@ -2068,9 +2071,20 @@ var caml_plonk_verifier_index_of_rust = function (x, affine_class) {
...
+      wasm_lookup_index,
+      poly_comm_class,
+      poly_comm_of_rust
+    );
+  }
+
   var caml_lookup_index = caml_opt_of_rust(
     x.lookup_index,
-    caml_lookup_verifier_index_of_rust
+    lookup_verifier_index_of_rust
```
mitschabaude, line 2087: this was passing `caml_lookup_verifier_index_of_rust` a function which expects multiple arguments (value and conversion helpers), to `caml_opt_of_rust` which call it with only the value argument

kimchi/js/bindings.js
```diff
@@ -2135,7 +2149,7 @@ var caml_lookup_patterns_to_rust = function (wasm_lookup_patterns, klass) {
   var range_check = wasm_lookup_patterns[3];
   var foreign_field_mul = wasm_lookup_patterns[4];
 
-  return new plonk_wasm.LookupPatterns(
+  return new plonk_wasm.WasmLookupPatterns(
```
mitschabaude, line 2152: this constructor previously didn't exist. It's now added to the Rust side along with two others

kimchi/wasm/src/plonk_verifier_index.rs
```diff
@@ -20,6 +20,68 @@ use std::path::Path;
...
+    pub fn new(max_per_row: usize, max_joint_size: u32, features: &WasmLookupFeatures) -> Self {
+        WasmLookupInfo {
+            inner: LookupInfo {
+                max_per_row,
+                max_joint_size,
+                features: features.inner.clone(),
+            },
+        }
+    }
+}
```
rbonichon, line 83: You should not have to do that for The WASM bindings for all of them are generated in `kimchi` (https://github.com/o1-labs/proof-systems/pull/1182 for develop, https://github.com/o1-labs/proof-systems/pull/1186 for berkeley)

kimchi/js/bindings.js
```diff
@@ -2159,12 +2173,11 @@ var caml_lookup_features_to_rust = function (wasm_lookup_features) {
...
-  var max_per_row = wasm_lookup_info[1];
-  var max_poly_size = wasm_lookup_info[2];
-  var features = caml_lookup_features_to_rust(wasm_lookup_info[3]);
-
-  return new plonk_wasm.LookupInfo(max_per_row, max_poly_size, features);
+var caml_lookup_info_to_rust = function (caml_lookup_info) {
+  var max_per_row = caml_lookup_info[1];
+  var max_poly_size = caml_lookup_info[2];
+  var features = caml_lookup_features_to_rust(caml_lookup_info[3]);
+  return new plonk_wasm.WasmLookupInfo(max_per_row, max_poly_size, features);
```
rbonichon, line 2180: Should be generated through `kimchi` as well

rbonichon (commented):
One of the key issue that I see is that you seem to be unable to use the code that should be generated from `proof-systems`.  Do you understand why that would be the case?

mitschabaude:
@rbonichon it's not true that we can't use code generated from proof-systems. This class definition code (in the compiled artifact) is generated from proof-systems:
image

This issue is that this contains no way to create instances of this class from that outside

mitschabaude:
By comparison, this is generated for the new class:
image

rbonichon:
`wires.rs` has the following

dannywillems:
Discussed with @rbonichon. He will update https://github.com/o1-labs/proof-systems/pull/1245/ to use `constructor` and we can therefore remove the `WasmCLASS` definitions.

dannywillems (approved):

PR #147, opened 2023-09-13
Update bindings for chunked zero-knowledge

mrmr1993:


kimchi/wasm/src/plonk_verifier_index.rs
```diff
@@ -744,7 +744,7 @@ macro_rules! impl_verification_key {
                         },
                     };
 
-                let (linearization, powers_of_alpha) = expr_linearization(Some(&feature_flags), true);
+                let (linearization, powers_of_alpha) = expr_linearization(Some(&feature_flags), true, 3);
```
rbonichon, line 747: What do you think about doing `let zk_rows = 3;` and using `zk_rows` everywhere there's a "magic" `3` occurring?

rbonichon (approved):
I suggested a minor improvement. Otherwise this is a straightforward change. LGTM.

mitschabaude:
Merging past the irrelevant prettier check, because the Mina part has been merged

PR #150, opened 2023-09-18
Fix Struct.toFields()

mitschabaude:
Remove unnecessary sorting so that the output of `Struct.toFields()` is always consistent with the output of `CircuitValue.toFields()` (So fields are output in the order in which their corresponding properties are defined).

Original PR by @Comdex: https://github.com/o1-labs/o1js-bindings/pull/145


ymekuria (approved):

Comdex:
@mitschabaude We may have missed some modifications. If I'm not mistaken, fromFields, toAuxiliary, toInput, toJSON and fromJSON in the provable method should all remove the sorting of object keys? 😅

https://github.com/o1-labs/o1js-bindings/blob/03e1b6ef52a3172aae15a92ea215d8b30bc83eee/lib/provable-snarky.ts#L47

mitschabaude:
Oh wow, great catch

PR #160, opened 2023-09-29
Bigint backend pt 8: Adapt conversion to latest kimchi types

mitschabaude:
Incremental step on top of #144 

Adapts TS kimchi types to latest OCaml / Rust code and introduces a (no-op) conversion of point evaluations which we need in #127 when switching to a different snarky field representation

### DIFF
--- a/crypto/bindings/conversion-proof.ts
+++ b/crypto/bindings/conversion-proof.ts
@@ -13,7 +13,6 @@ import type {
 import type * as wasmNamespace from '../../compiled/node_bindings/plonk_wasm.cjs';
 import type {
   OrInfinity,
-  LookupEvaluations,
   PointEvaluations,
   PolyComm,
   ProverProof,
@@ -162,7 +161,7 @@ function proofConversionPerField(
       let commitments = commitmentsToRust(proof[1]);
       let openingProof = openingProofToRust(proof[2]);
       // TODO typed as `any` in wasm-bindgen, this has the correct type
-      let evals: ProofEvaluations<Uint8Array> = proof[3];
+      let evals = proofEvaluationsToRust(proof[3]);
       let ftEval1 = fieldToRust(proof[4]);
       let public_ = fieldsToRustFlat(proof[5]);
       let [, ...prevChallenges] = proof[6];
@@ -187,7 +186,10 @@ function proofConversionPerField(
     proofFromRust(proof: WasmProverProof): ProverProof {
       let commitments = commitmentsFromRust(proof.commitments);
       let openingProof = openingProofFromRust(proof.proof);
-      let evals: ProofEvaluations<Uint8Array> = proof.evals;
+      let evals = proofEvaluationsFromRust(
+        // TODO typed as `any` in wasm-bindgen, this is the correct type
+        proof.evals satisfies ProofEvaluations<Uint8Array>
+      );
       let ftEval1 = fieldFromRust(proof.ft_eval1);
       let public_ = fieldsFromRustFlat(proof.public_);
       let prevChallengeScalars = proof.prev_challenges_scalars;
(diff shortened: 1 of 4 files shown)


crypto/bindings/conversion-proof.ts
```diff
@@ -187,7 +186,10 @@ function proofConversionPerField(
     proofFromRust(proof: WasmProverProof): ProverProof {
       let commitments = commitmentsFromRust(proof.commitments);
       let openingProof = openingProofFromRust(proof.proof);
-      let evals: ProofEvaluations<Uint8Array> = proof.evals;
+      let evals = proofEvaluationsFromRust(
+        // TODO typed as `any` in wasm-bindgen, this is the correct type
+        proof.evals satisfies ProofEvaluations<Uint8Array>
```
dannywillems, line 191: Can't we get rid of this? Why wasm-bindgen type it as any?

crypto/bindings/conversion-proof.ts
```diff
@@ -222,32 +224,68 @@ function mapProofEvaluations<Field1, Field2>(map: (x: Field1) => Field2) {
...
+  const mapPointEvalsOption = (
+    evals: MlOption<PointEvaluations<Field1>>
+  ): MlOption<PointEvaluations<Field2>> => MlOption.map(evals, mapPointEvals);
 
   return function mapProofEvaluations(
     evals: ProofEvaluations<Field1>
   ): ProofEvaluations<Field2> {
-    let [, w, z, s, coeffs, lookup, genericSelector, poseidonSelector] = evals;
+    let [
+      ,
```
dannywillems, line 235: There is now the public evaluations in develop.

crypto/bindings/conversion-proof.ts
```diff
@@ -222,32 +224,68 @@ function mapProofEvaluations<Field1, Field2>(map: (x: Field1) => Field2) {
...
+      rotSelector,
+      lookupAggregation,
+      lookupTable,
+      lookupSorted,
+      runtimeLookupTable,
+      runtimeLookupTableSelector,
+      xorLookupSelector,
+      lookupGateLookupSelector,
+      rangeCheckLookupSelector,
+      foreignFieldMulLookupSelector,
```
dannywillems, line 260: LGTM, checked with https://github.com/o1-labs/proof-systems/commit/4a87b0831c6230b572705f25eb664ca28e778bd4

crypto/bindings/conversion-verifier-index.ts
```diff
@@ -184,20 +184,29 @@ function verifierIndexConversionPerField(
   function lookupSelectorsToRust([
     ,
     lookup,
```
dannywillems, line 186: lookup seems to come first in Rust

crypto/bindings/conversion-verifier-index.ts
```diff
@@ -184,20 +184,29 @@ function verifierIndexConversionPerField(
...
     let lookup = MlOption.mapTo(selector.lookup, core.polyCommFromRust);
+    let xor = MlOption.mapTo(selector.xor, core.polyCommFromRust);
+    let range_check = MlOption.mapTo(
+      selector.range_check,
+      core.polyCommFromRust
+    );
+    let ffmul = MlOption.mapTo(selector.ffmul, core.polyCommFromRust);
     selector.free();
-    return [0, lookup];
+    return [0, lookup, xor, range_check, ffmul];
```
dannywillems, line 209: xor comes first

crypto/bindings/lookup.ts
```diff
@@ -21,7 +21,13 @@ type LookupInfo = [
   max_joint_size: number,
   features: LookupFeatures
 ];
-type LookupSelectors<PolyComm> = [_: 0, lookup: MlOption<PolyComm>];
+type LookupSelectors<PolyComm> = [
+  _: 0,
+  lookup: MlOption<PolyComm>,
```
dannywillems, line 26: xor comes first

dannywillems (changes requested):
Type is wrong for lookup selectors. Also might require changes to be compatible with develop.

crypto/bindings/lookup.ts
mitschabaude: @dannywillems this has to match the type here, where `xor` comes second: https://github.com/MinaProtocol/mina/blob/7a975d31aee7efef5cadc215ea0386a02bdf8808/src/lib/crypto/kimchi_bindings/stubs/kimchi_types.ml#L200-L205

mitschabaude:
Hey @dannywillems this shouldn't be compatible with `develop` but with `mina/snarkyjs-main` (which is compatible with the current testnet). I already brought it up to date with develop end of last week in a different branch: https://github.com/o1-labs/o1js-bindings/pull/161

crypto/bindings/lookup.ts
dannywillems: Oh :o ... Good catch. Let's open an issue to fix the wrong order, just in case we make some assumptions in the order somewhere.

crypto/bindings/conversion-proof.ts
mitschabaude: I don't know... here is the type definition on the Rust side: https://github.com/o1-labs/o1js-bindings/blob/d185e8609205aa820eb01c7317e4b48e3c0d7d53/kimchi/wasm/src/plonk_proof.rs#L474-L476

dannywillems: Let's leave it like this for the moment. Opening followup issue.

dannywillems (approved):

Issue #166, opened 2023-10-03
Apple M2 cpus not optimized


MartinOndejka:
[Fix here](https://github.com/o1-labs/o1js-bindings/blob/main/js/node/node-backend.js#L144) doesn't catch all of the M2 cpus, like `Apple M2 Pro` etc. I would fix it myself, but I don't know where did those numbers come from.

mitschabaude:
Oh nice catch @MartinOndejka! The numbers came from testing which number of CPUs was the fastest

mitschabaude:
Feel free to test proof generation on your machine and put the best-performing number there!

jackryanservia:
Fixed permanently in o1-labs/o1js#1456 :)

PR #172, opened 2023-10-05
Add other M2 variations to the list of apple silicon cpus

MartinOndejka:
Anyone with access to the M2 Ultra/Max feel free to suggest what should be the best number of workers.

https://github.com/o1-labs/o1js-bindings/issues/166


mitschabaude (approved):
looks like a reasonable guess :D

PR #174, opened 2023-10-06
Remove the sorting of Object keys in Struct related methods

Comdex:
This is a complement to the previous PR #150 , keeping the treatment consistent.


mitschabaude (approved):
thanks for adding this!

PR #176, opened 2023-10-09
Merge main and adapt TS bindings to develop

mitschabaude:
* merge `main` into `develop`
* adapt the [bindings refactor](https://github.com/o1-labs/o1js-bindings/pull/127) to `develop`

o1js: https://github.com/o1-labs/o1js/pull/1168


mitschabaude:
The linter checks seem to be using the wrong prettier oncfig somehow. Locally, running `npx prettier -c kimchi/js/**/*.js js/**/*.js` works

crypto/bindings/conversion-verifier-index.ts
```diff
@@ -115,6 +115,12 @@ function verifierIndexConversionPerField(
     let mulComm = core.polyCommToRust(evals[6]);
     let emulComm = core.polyCommToRust(evals[7]);
     let endomulScalarComm = core.polyCommToRust(evals[8]);
+    let xorComm = MlOption.mapFrom(evals[9], core.polyCommToRust);
```
dannywillems, line 118: xorComm and rangeCheck0Comm are the same

dannywillems:
245 commits...
image
Source: https://www.youtube.com/watch?v=QglFGVDcuX8

crypto/bindings/conversion-proof.ts
```diff
@@ -186,6 +224,7 @@ function proofConversionPerField(
       let prevChallengeScalars = new VecVec(n);
       let prevChallengeCommsMl: MlArray<PolyComm> = [0];
       for (let [, scalars, comms] of prevChallenges) {
+        // TODO is .push() correct if we already initialized to the final length?
```
dannywillems, line 243: Is there a test to verify it?

crypto/bindings/conversion-proof.ts
```diff
@@ -173,6 +189,28 @@ function proofConversionPerField(
...
+
+  function lookupTableToRust([
+    ,
+    id,
+    [, ...data],
+  ]: LookupTable): WasmLookupTable {
+    let n = data.length;
+    let wasmData = new VecVec(n);
+    for (let i = 0; i < n; i++) {
+      // TODO is .push() correct if we already initialized to the final length?
```
dannywillems, line 221: Same than below.

kimchi/js/bindings.js
```diff
@@ -414,7 +414,7 @@ var caml_pasta_fq_plonk_circuit_serialize = function (
 };
 
 // Provides: caml_pasta_fp_plonk_index_create
-// Requires: plonk_wasm, free_on_finalize, tsRustConversion
+// Requires: plonk_wasm, tsRustConversion, free_on_finalize
```
dannywillems, line 417: Useless change.

kimchi/js/bindings.js
```diff
@@ -516,7 +516,7 @@ var caml_pasta_fp_plonk_index_write = function (append, t, path) {
 };
 
 // Provides: caml_pasta_fq_plonk_index_create
-// Requires: plonk_wasm, free_on_finalize, tsRustConversion
+// Requires: plonk_wasm, tsRustConversion, free_on_finalize
```
dannywillems, line 519: Useless change.

scripts/update-snarkyjs-bindings.sh
```diff
@@ -12,11 +12,14 @@ WEB_BINDINGS="$SNARKY_JS_PATH/src/bindings/compiled/web_bindings"
 
 # 1. node build
 
-$DIR_PATH/build-snarkyjs-node.sh
+$DIR_PATH/build-snarkyjs-node-artifacts.sh
```
dannywillems, line 15: ? Why do you change this?

crypto/bindings/conversion-proof.ts
```diff
@@ -50,6 +51,14 @@ export { proofConversion };
 
 const proofEvaluationsToRust = mapProofEvaluations(fieldToRust);
 const proofEvaluationsFromRust = mapProofEvaluations(fieldFromRust);
+const pointEvalsOptionToRust = mapPointEvalsOption(fieldToRust);
+const pointEvalsOptionFromRust = mapPointEvalsOption(fieldFromRust);
+
+type WasmProofEvaluations = [
+  0,
+  MlOption<PointEvaluations<Uint8Array>>,
+  ...RemoveLeadingZero<ProofEvaluations<Uint8Array>>
```
dannywillems, line 60: ? Where this type come from `RemoveLeadingZero`?

dannywillems: Oh I see:
```
type RemoveLeadingZero<T extends [0, ...any]> = T extends [0, ...infer U]
  ? U
  : never;
```

crypto/bindings/conversion-proof.ts
```diff
@@ -297,13 +312,20 @@ function proofConversionPerField(
   };
 }
 
-function mapProofEvaluations<Field1, Field2>(map: (x: Field1) => Field2) {
-  const mapPointEvals = (
-    evals: PointEvaluations<Field1>
-  ): PointEvaluations<Field2> => {
+function createMapPointEvals<Field1, Field2>(map: (x: Field1) => Field2) {
```
dannywillems, line 315: Why `Field1` and `Field2`?

scripts/update-snarkyjs-bindings.sh
mitschabaude: the existing script failed to update the bindings when the TS compilation didn't work. but now some of our TS depends on the wasm-bindgen types. so there was sometimes a circular dependency were you could never compile. this change fixes it (compiling Rust bindings doesn't need TS to succeed)

mitschabaude:
> > The linter checks seem to be using the wrong prettier config somehow. Locally, running `npx prettier -c kimchi/js/**/*.js js/**/*.js` works
> 
> Did you try without `npx`? Is it not gonna get the prettier file where it finds the package.json file? In anycase, it must be fixed.

I think the CI is wrong, not my local setup :D I can see that what my auto-formatter does corresponds to the prettier config we have (single quotes)

crypto/bindings/conversion-proof.ts
mitschabaude: the point of these functions is to map a data structure containing bigint fields (used by TS and snarky) to one containing `Uint8Array` fields (used by Rust). it's generic so that we can use it to map both ways

crypto/bindings/conversion-verifier-index.ts
mitschabaude: fixed in subsequent commit

dannywillems:
Something important to check is if proof_create and index_create has a bytecode version.

crypto/bindings/conversion-proof.ts
mitschabaude: I now read the Rust docs for `Vec::with_capacity` and it's clear from there that this is correct

mitschabaude: removing the TODO comments

mitschabaude:
> Something important to check is if proof_create and index_create has a bytecode version.

Confirmed that this is the case 👍🏻

mitschabaude:
Fixed the lint job @dannywillems, see last commit ^^

dannywillems (approved):

dannywillems (approved):

PR #177, opened 2023-10-10
64-bit range check gate

mitschabaude:
- expose range_check0 gate
- adapt mli


mitschabaude:
superseded by merged PR

PR #179, opened 2023-10-11
Activate custom gates in JS and add rangeCheck64 gadget

mitschabaude:
companion of https://github.com/o1-labs/o1js/pull/1176, see PR description there


kimchi/wasm/src/plonk_verifier_index.rs
```diff
@@ -20,6 +20,41 @@ use std::path::Path;
 use std::sync::Arc;
 use wasm_bindgen::prelude::*;
 
+pub struct SimpleFeatureFlags {
+    pub range_check0: bool,
+    pub range_check1: bool,
```
MartinMinkov, line 25: Just a clarifying question for me: what does `range_check1` correspond to? `range_check0` is the `64` bit range check; what is the other one?

mitschabaude: Here's how I understand it (@jspada would know the full details):

`range_check0` actually has space for up to 88 bits (for the 64-bit check we set the upper two 12-bit limbs to constant zero). However, when used in the 88-bit mode it doesn't fully constrain all limbs by itself and needs to "collaborate" with subsequent gates.

The main objective for both range check gates was to enable a 3x88-bit check (used in foreign field arithmetic, supports foreign field element sizes up to 264 = 3 * 88 bits). The `range_check1` gate is what is needed together with two `range_check0` gates and one `Zero` gate to collaboratively do this 3*88 bit check.

Note that that's 264 bits of range check in 4 gates, which is 264 / 4 = 66 bits per gate on average. So the collaborative approach gives you a 2 bit edge over the single gate version here 😅 Just doing 4 64-bit range checks separately would have only given us 265 bit checks which would've been to small for important use cases in foreign field multiplication.

mitschabaude: check out https://github.com/MinaProtocol/mina/blob/e2b96abe89faf0caa92a75495b9b4b0cb96f2d58/src/lib/crypto/kimchi_backend/gadgets/range_check.ml

ocaml/lib/snarky_bindings.mli
```diff
@@ -51,6 +51,23 @@ val snarky :
           Js.meth >
       Js.t
       Js.readonly_prop
+  ; gates :
+      < rangeCheck0 :
+          (   Field.t
```
dannywillems, line 56: I would be in favor of documenting a bit the parameters as we are in an interface file :)

ocaml/lib/pickles_bindings.mli
```diff
@@ -30,6 +30,7 @@ type pickles_rule_js =
           ; shouldVerify : Boolean.var array Js.prop >
           Js.t )
       Js.prop
+  ; featureFlags : bool Pickles_types.Plonk_types.Features.t Js.prop
```
dannywillems, line 33: I would be in favor of documenting a bit the parameters as we are in an interface file :)

kimchi/wasm/src/plonk_verifier_index.rs
```diff
@@ -20,6 +20,41 @@ use std::path::Path;
 use std::sync::Arc;
 use wasm_bindgen::prelude::*;
 
+pub struct SimpleFeatureFlags {
```
dannywillems, line 23: I'm not totally in favor of adding a new layer of "complexity" with a new structure, which is only used to build `FeatureFlags`. I would suggest either we simplify `FeatureFlags` or add a new constructor or we remove this layer.

mitschabaude: Ok I can just move this logic into the function that computes the feature flags below. Custom constructor for FeatureFlags is probably the way to go. Note that pickles uses a representation equivalent to SimpleFeatureFlags

ocaml/lib/pickles_bindings.mli
mitschabaude: I'm not sure if this logic applies here, because this interface is only used from JS, not from OCaml, so the comments won't enhance any editor experience. Probably well documented interfaces on the JS side are a better investment

dannywillems: Regarding the interface, it will still be used by the o1js maintainer.

mitschabaude: addressed here instead because this is where the interface is exposed: https://github.com/o1-labs/o1js/pull/1176/commits/68b9ba89d4daeeb59f5e93e65c557be3d6d96328

PR #180, opened 2023-10-11
expose xor gate

Trivo25:
o1js https://github.com/o1-labs/o1js/pull/1177


crypto/finite_field.ts
```diff
@@ -123,6 +123,9 @@ function createField(p: bigint, t: bigint, twoadicRoot: bigint) {
     t,
     twoadicRoot,
 
+    xor(x: bigint, y: bigint) {
+      return y ^ x;
+    },
```
mitschabaude, line 128: what I don't love about this function is that the output is not necessarily another field element (i.e. smaller than the modulus)

it seems a bit out of place in a finite field library. maybe we need a bigint library as well. maybe for another PR.

ocaml/lib/snarky_bindings.ml
```diff
@@ -141,6 +141,43 @@ module Gates = struct
...
+                   ; in2_3
+                   ; out_0
+                   ; out_1
+                   ; out_2
+                   ; out_3
+                   } )
+          } )
+
+  let zero_check in1 in2 out =
+    Impl.with_label "zero_check" (fun () ->
```
mitschabaude, line 172: as discussed in the other PR this should probably just be called `zero`

mitschabaude (approved):
pre-approving with a nit

crypto/finite_field.ts
mitschabaude: sorry for being annoying but I don't like that either 😅 that behavior is unexpected. it's fine to leave the function here for now but better remove the `mod()`

Trivo25: That's fine, I don't mind at all! I agree its an odd place for it to live in :D

PR #184, opened 2023-10-12
Add rot function to bindings

MartinMinkov:
# Summary

This PR exposes the `ROT` constraint to our JS bindings and adds a `rot` function for finite fields in the case of constants.

## Context
In the context of bitwise operations, a rotation (often referred to as a "bitwise rotation") is an operation that shifts the bits of a binary number either to the left or to the right, but unlike a standard shift operation, the bits that "fall off" the end are not discarded. Instead, they "wrap around" to the other end.

There are two primary types of rotations:

1. Rotate Left (ROL): In this operation, bits are shifted to the left. The bits that fall off the leftmost side "wrap around" and reappear on the rightmost side.
2. Rotate Right (ROR): In this operation, bits are shifted to the right. The bits that fall off the rightmost side "wrap around" and reappear on the leftmost side.

# Tested
This work is tested in the upper layers of `o1js` with unit and e2e tests that are built on top of the exposed gate constraint.

:link: o1js: https://github.com/o1-labs/o1js/pull/1182


crypto/finite_field.ts
```diff
@@ -178,5 +178,22 @@ function createField(p: bigint, t: bigint, twoadicRoot: bigint) {
     fromBigint(x: bigint) {
       return mod(x, p);
     },
+    rot(x: bigint, bits: number, direction: boolean = true) {
```
mitschabaude, line 181: can we make `direction: 'left' | 'right'` so that it's clear what the parameter means without looking at the function code?

mitschabaude (approved):
pre-approving with the comment above!

crypto/finite_field.ts
MartinMinkov: done!

PR #187, opened 2023-10-19
Cache prover keys

mitschabaude:
Companion to https://github.com/o1-labs/o1js/pull/1187

* Set Pickles' `cache` and `storables` arguments to values wrapping a `storage` argument provided from the JS side
  * this enables caching of prover and verifier keys from JS
  * `storables` argument is added to Pickles in the Mina companion: https://github.com/MinaProtocol/mina/pull/14396
* Expose bindings functions to encode and decode prover keys to byte arrays


ocaml/lib/util.ml
```diff
@@ -1,6 +1,25 @@
...
+module Js_environment = struct
+  type t = Node | Web | Unknown
+
+  let value =
+    let env = get_ts_bindings () |. "jsEnvironment" in
+    Js.(
+      if phys_equal env (def (string "node")) then Node
+      else if phys_equal env (def (string "web")) then Web
+      else Unknown)
+end
```
mitschabaude, line 21: this makes the ocaml bindings aware of the current JS environment. I originally thought that I needed this to avoid using nodejs filesystem operations on the web. in the design I ended up with, the `write` and `read` operations are generic and can be set to something which works in either nodejs or the web on the JS side; so this remains unused.
It still seems useful to have.

kimchi/wasm/src/plonk_verifier_index.rs
```diff
@@ -910,9 +910,12 @@ macro_rules! impl_verification_key {
...
                 index: String,
-            ) -> WasmPlonkVerifierIndex {
-                let vi: DlogVerifierIndex<$G, OpeningProof<$G>> = serde_json::from_str(&index).unwrap();
-                return to_wasm(srs, vi.into())
+            ) -> Result<WasmPlonkVerifierIndex, JsError> {
+                let vi: Result<DlogVerifierIndex<$G, OpeningProof<$G>>, serde_json::Error> = serde_json::from_str(&index);
+                match vi {
+                    Ok(vi) => Ok(to_wasm(srs, vi)),
+                    Err(e) => Err(JsError::new(&(e.to_string()))),
+                }
```
mitschabaude, line 918: verification key encoding was already in place in a usable way (encode as JSON string, this is fine for vks which are small). I just changed the deserialization so that it can throw a proper error and not just panic

crypto/bindings.ts
```diff
@@ -51,3 +55,9 @@ function createRustConversion(wasm: wasm) {
     mapMlArrayToRustVector: core.mapMlArrayToRustVector,
   };
 }
+
+let rustConversion: ReturnType<typeof createRustConversion> | undefined;
+
+function getRustConversion(wasm: wasm) {
+  return rustConversion ?? (rustConversion = createRustConversion(wasm));
+}
```
mitschabaude, line 63: this is exposed to the main o1js codebase

ocaml/lib/pickles_bindings.ml
```diff
@@ -620,4 +787,15 @@ let pickles =
       fun (proof : proof) ->
         proof |> Pickles.Side_loaded.Proof.of_proof
         |> Pickles.Side_loaded.Proof.to_base64 |> Js.string
+
+    val encodeVerificationKey = encode_verification_key
+
+    val decodeVerificationKey = decode_verification_key
+
+    val util =
```
dannywillems, line 795: Do you use this?

ocaml/lib/pickles_bindings.ml
```diff
@@ -602,12 +744,37 @@ let dummy_verification_key () =
...
+    @@ Js.to_string bytes
+  in
+  let open Ppx_deriving_yojson_runtime.Result in
+  match vk_or_error with
+  | Ok vk ->
+      vk
+  | Error err ->
+      failwithf "Could not decode verification key: %s" err ()
+
+module Util = struct
```
dannywillems, line 762: It seems to be a simple alias. Is it really useful?

ocaml/lib/pickles_bindings.ml
```diff
@@ -368,6 +369,134 @@ module Choices = struct
...
+            match res with Error e -> Some e | Ok () -> None )
+      in
+      match errs with [] -> Ok () | errs -> Error (Error.of_list errs)
+  end
+
+  let () = Key_cache.set_sync_implementation (module Sync)
+
+  open Pickles.Cache
+
+  type any_key =
```
dannywillems, line 418: That seems to be code that can be refactorised, but I don't have any idea off the top of my head.

dannywillems (approved):

crypto/bindings/env.ts
```diff
@@ -0,0 +1 @@
+export const jsEnvironment = 'node';
```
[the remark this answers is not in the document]
mitschabaude: No, I can explain why -- the `.web.ts` extension thing is a custom feature of our build process which is implemented as part of the web build (it looks for `.web.ts` and replaces the `.ts` versions)

However, we need the normal `.ts` extension for the TS compilation. There's no way to configure TS (in ESM mode) to look for a different file than what is imported. Also, the Node.js build is very simple and would be complicated by considering special extensions.

ocaml/lib/pickles_bindings.ml
mitschabaude: I could put it in the `js%object` declaration, that's true. currently there is a consistent style of not putting any logic in js object declaration, and _only_ using aliases there. in this case I could make an exception

ocaml/lib/pickles_bindings.ml
mitschabaude: yes, in one place, here: https://github.com/o1-labs/o1js/pull/1187/files#diff-a0ca705d913f22307bcab0d9758200d95493afcd46279320867493dada7bce68R56

ocaml/lib/pickles_bindings.ml
mitschabaude: we could pass a separate storage implementation from JS for every type of prover/verifier key -- then we wouldn't need those enums. but it would make the OCaml-TS interface much bigger. I prefer keeping the interface as small as possible because it is the most brittle part of the system (no type guarantees etc), that's why I went with a single `storable` that is passed from JS, which handles all key types at once

PR #190, opened 2023-10-23
Expose Basic Gate (for AND gadget)

jackryanservia:
# Summary
This PR exposes the generic constraint to our JS bindings via a function called `Basic` and is part of the work necessary for the AND gadget.

## Context
The AND gadget behaves similarly to the `&` operator in JavaScript and does not require a new custom gate. Instead, it uses the [XOR gate](https://github.com/o1-labs/o1js/pull/1177) and one generic gate. See [OCaml implementation](https://o1-labs.github.io/proof-systems/specs/kimchi.html?highlight=gate#and) for context.

## Tested
This work is tested in the upper layers of o1js with unit and e2e tests built on top of the exposed gate constraint.

AND gadget PR in o1js repo: https://github.com/o1-labs/o1js/pull/1193

closes: https://github.com/o1-labs/o1js/issues/1140


ocaml/lib/snarky_bindings.ml
```diff
@@ -178,6 +178,15 @@ module Gates = struct
                    { kind = Zero; values = [| in1; in2; out |]; coeffs = [||] }
                 )
           } )
+
+  let basic sl l sr r so o sm sc =
```
jackryanservia, line 182: I assumed that it made sense to expose the basic gate and do all of the configuration in TS instead of exposing `and_equation` as in the [OCaml implementation](https://github.com/MinaProtocol/mina/blob/d5ae586319a9d0fdfd322b2b8b54beb5bfaeddbb/src/lib/crypto/kimchi_backend/gadgets/bitwise.ml#L476). 
1. Is this a correct assumption?
1. Are the argument names good?
1. Is there anything else I should expose while I'm here?

jackryanservia (commented):
I think everything is good, but I left a few questions :).

ocaml/lib/snarky_bindings.ml
jackryanservia, line 182: I assumed that it made sense to expose the basic gate and do all of the configuration in TS instead of exposing `and_equation` as in the [OCaml implementation](https://github.com/MinaProtocol/mina/blob/d5ae586319a9d0fdfd322b2b8b54beb5bfaeddbb/src/lib/crypto/kimchi_backend/gadgets/bitwise.ml#L476). 
1. Is this a correct assumption?
1. Are the argument names good?
1. Is there anything else I should expose while I'm here?

jackryanservia, line 182: I assumed that it made sense to expose the basic gate and do all of the configuration in TS instead of exposing `and_equation` as in the [OCaml implementation](https://github.com/MinaProtocol/mina/blob/d5ae586319a9d0fdfd322b2b8b54beb5bfaeddbb/src/lib/crypto/kimchi_backend/gadgets/bitwise.ml#L476). 
1. Is this a correct assumption?
1. Are the argument names good?
1. Is there anything else I should expose while I'm here?

ocaml/lib/snarky_bindings.ml
```diff
@@ -178,6 +178,15 @@ module Gates = struct
...
+    Impl.with_label "generic_gate" (fun () ->
+        Impl.assert_
+          { annotation = Some __LOC__
+          ; basic =
+              Kimchi_backend_common.Plonk_constraint_system.Plonk_constraint.T
+                (Basic { l = (sl, l); r = (sr, r); o = (so, o); m = sm; c = sc })
+          } )
```
jackryanservia, line 189: I exposed `Basic` and configured the gate in TS insetead of implementing `and_equation` as in the [OCaml code](https://github.com/MinaProtocol/mina/blob/d5ae586319a9d0fdfd322b2b8b54beb5bfaeddbb/src/lib/crypto/kimchi_backend/gadgets/bitwise.ml#L476).

1. Is this the right approach?
1. Are the argument names appropriate?
1. Is there anything else I should expose while I'm here?

crypto/finite_field.ts
```diff
@@ -126,6 +126,9 @@ function createField(p: bigint, t: bigint, twoadicRoot: bigint) {
     xor(x: bigint, y: bigint) {
       return x ^ y;
     },
+    and(x: bigint, y: bigint) {
+      return x & y;
+    },
```
jackryanservia, line 131: It looks like [we'll pull this out into its own library](https://github.com/o1-labs/o1js-bindings/pull/180#discussion_r1362285209) sometime in the future, but I assume this is alright for now?

ocaml/lib/snarky_bindings.ml
mitschabaude: 1a) Yes it's the right approach to expose the generic gate in it's most general form and not specialized to the "and" logic. We definitely want this in o1js, out of principle!
1b) I don't think we need it for the "and" gadget though. o1js creates the right generic gates when you do `.add()` and `mul()` and so on, which we could use for "and" (will comment on the other PR)
2) I think so. I guess `sl` means "the selector for the left input" etc?
3) No I think this is good, although I think we should out of principle make an effort to expose _all_ gates to o1js, but that can be a separate PR

ocaml/lib/snarky_bindings.ml
```diff
@@ -358,6 +367,8 @@ let snarky =
         method xor = Gates.xor
 
         method zero = Gates.zero
+
+        method basic = Gates.basic
```
mitschabaude, line 371: Can we call the gate `generic` instead of `basic`? At least on the TS side

mitschabaude (approved):
Approving with a small renaming ask

crypto/finite_field.ts
jackryanservia: Oh! That makes a lot of sense 😂 ! Should I eliminate this and do it all in TS with the infix operator (I could also update Florian's implementation), or should we just leave it? (I don't mind changing!)

mitschabaude: I vote yes!

ocaml/lib/snarky_bindings.ml
jackryanservia: 1b) Ahhhh! My assumption was that that would be less efficient than doing what I did, but I think I understand now. I gather from Anaïs that when I call `.add()`, it goes on to a stack, and then once there is enough on the stack for us to generate an efficient constraint we do so. Is that correct? (Don't worry about tying it all out here if not; maybe we can just do a short call sometime and I will make a list of questions).
2) Correct!

mitschabaude: Yeah what Anais wrote is true but she probably didn't know that a simple `Field.mul()` also flushes that stack into a constraint, so we don't need that complicated way of specifying the gate

PR #191, opened 2023-10-23
Add left shift and right shift to finite field lib

MartinMinkov:
# Summary
This PR  adds left and right shift functions for finite fields in the case of constants.

# Tested
This work is tested in the upper layers of o1js with unit and e2e tests that are built on top of the exposed gate constraint.

:link: o1js: https://github.com/o1-labs/o1js/pull/1194


crypto/finite_field.ts
```diff
@@ -195,5 +195,29 @@ function createField(p: bigint, t: bigint, twoadicRoot: bigint) {
       }
       return BigInt('0b' + binary.reverse().join(''));
     },
+    leftShift(x: bigint, bits: number) {
```
MartinMinkov, line 198: Unit tests in `o1js` were failing when `leftShift` was defined like this:

```ts
leftShift(x: bigint, bits: number) {
    return mod(x << BigInt(bits), p);
}
```

I'm wondering if the reviewer knows why this is the case? I had to define a manual shift implementation instead to get unit tests to pass.

Trivo25, line 198: why not just `x << BigInt(5)`?

Trivo25: Its probably because the shift operator in JavaScript (on `bigint`) does not make a "cut" after 64 bit and just keeps shifting. I believe bitwise operations on `number` are limited to 31bit, `bigint` isn't

Trivo25:
```ts
leftShift(1n, 64)
> 0
1n << 65n
> 36893488147419103232
```

MartinMinkov: Fixed in https://github.com/o1-labs/o1js-bindings/pull/191/commits/169c97fa2ab1ef1ee345496795a067756fd8eab2

crypto/finite_field.ts
```diff
@@ -195,5 +195,12 @@ function createField(p: bigint, t: bigint, twoadicRoot: bigint) {
       }
       return BigInt('0b' + binary.reverse().join(''));
     },
+    leftShift(x: bigint, bits: number, maxBitSize: number = 64) {
+      let shifted = x << BigInt(bits);
+      return mod(BigInt('0b' + shifted.toString(2).slice(-maxBitSize)), p);
```
MartinMinkov, line 200: not sure if doing the `mod` here is needed. To the reviewer, should this be removed?

Trivo25: it wont do anything, assuming that `maxBitSize` is always `maxBitSize <= p`, but I start to agree with Gregor that these functions shouldnt belong in the `finite_field` library. I would remove `mod` and in a future PR clean up all the bitwise operations so that they dont introduce "impurities" in the ff lib :D

Trivo25 (approved):

crypto/finite_field.ts
```diff
@@ -195,5 +195,12 @@ function createField(p: bigint, t: bigint, twoadicRoot: bigint) {
       }
       return BigInt('0b' + binary.reverse().join(''));
     },
+    leftShift(x: bigint, bits: number, maxBitSize: number = 64) {
+      let shifted = x << BigInt(bits);
+      return BigInt('0b' + shifted.toString(2).slice(-maxBitSize));
```
mitschabaude, line 200: nit: more efficient would be
```suggestion
      return shifted & ((1n << BigInt(maxBitSize)) - 1n);
```

PR #192, opened 2023-10-25
SRS caching

mitschabaude:
Cache SRS and Lagrange basis

This
* relies on https://github.com/o1-labs/proof-systems/pull/1300 to create SRS in parallel
* moves bindings implementation of srs and lagrange commitments creation to TS
* in the new implementation, supports user-definable cache to read/write srs and lagrange basis for different domain sizes


kimchi/wasm/src/srs.rs
```diff
@@ -213,35 +215,171 @@ macro_rules! impl_srs {
...
-        Vesta,
-        WasmPolyComm,
-        Fp
-    );
+    use mina_curves::pasta::{Fp, Vesta as G};
+    use poly_commitment::PolyComm;
+
+    impl_srs!(caml_fp_srs, WasmPastaFp, WasmG, Fp, G, WasmPolyComm, Fp);
+
+    #[wasm_bindgen]
```
dannywillems, line 228: It seems it is duplicated code that can be parametrized by the curve (I might be wrong by reading only GitHub panel). Can it be moved into the macro `impl_srs!`?

scripts/update-wasm-and-types.sh
```diff
@@ -0,0 +1,31 @@
...
+pushd ../../..
+  dune b $KIMCHI_BINDINGS/js/node_js
+
+  chmod 777 "$_NODE_BINDINGS"/*
+
+  cp _build/default/$KIMCHI_BINDINGS/js/node_js/plonk_wasm* "$_NODE_BINDINGS"/
+  mv -f $_NODE_BINDINGS/plonk_wasm.js $_NODE_BINDINGS/plonk_wasm.cjs
+  mv -f $_NODE_BINDINGS/plonk_wasm.d.ts $_NODE_BINDINGS/plonk_wasm.d.cts
+
+  chmod 777 "$_NODE_BINDINGS"/*
```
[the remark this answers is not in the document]
mitschabaude: something in the script creates files with different permissions. I think it's JSOO. so I have to change the permissions, otherwise it won't allow me to copy the files

scripts/update-wasm-and-types.sh
```diff
@@ -0,0 +1,31 @@
...
+
+set -e
+
+SNARKY_JS_PATH="src/lib/snarkyjs"
+DIR_PATH=$(dirname "$0")
+KIMCHI_BINDINGS="$SNARKY_JS_PATH/src/bindings/kimchi"
+_NODE_BINDINGS="$SNARKY_JS_PATH/src/bindings/compiled/_node_bindings"
+NODE_BINDINGS="$SNARKY_JS_PATH/src/bindings/compiled/node_bindings"
+
+export DUNE_USE_DEFAULT_LINKER="y"
```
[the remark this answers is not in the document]
mitschabaude: I don't know 😬 Matthew put it in the main o1js build script, and I copied from there

kimchi/wasm/src/srs.rs
mitschabaude: you're right, could be in the macro. I think it's easier to maintain outside the macro, see my slack thread here: https://o1-labs.slack.com/archives/C01UNQH8XEZ/p1698141695847899

crypto/bindings/srs.ts
```diff
@@ -0,0 +1,215 @@
...
+            if (cache.canWrite) {
+              let wasmSrs = getSrs(srs);
+
+              let mlSrs = conversion[f].pointsFromRust(wasmSrs);
+              let jsonSrs = MlArray.mapFrom(mlSrs, OrInfinity.toJSON);
+              let bytes = new TextEncoder().encode(JSON.stringify(jsonSrs));
+
+              try {
+                cache.write(header, bytes);
+              } catch (e) {}
```
dannywillems, line 123: we don't log the error?

kimchi/wasm/src/srs.rs
```diff
@@ -75,10 +75,12 @@ macro_rules! impl_srs {
                 srs: &[<Wasm $field_name:camel Srs>],
                 log2_size: i32,
             ) {
-                let ptr: &mut poly_commitment::srs::SRS<$G> =
-                    unsafe { &mut *(std::sync::Arc::as_ptr(&srs) as *mut _) };
-                let domain = EvaluationDomain::<$F>::new(1 << (log2_size as usize)).expect("invalid domain size");
-                ptr.add_lagrange_basis(domain);
+                crate::rayon::run_in_pool(|| {
```
Trivo25, line 78: looks like a simple but powerful change lol

crypto/bindings/srs.ts
```diff
@@ -0,0 +1,215 @@
...
+            // TODO: this takes a bit to long, about 300ms for 2^16
+            // `pointsToRust` is the clear bottleneck
+            let jsonSrs: OrInfinityJson[] = JSON.parse(
+              new TextDecoder().decode(bytes)
+            );
+            let mlSrs = MlArray.mapTo(jsonSrs, OrInfinity.fromJSON);
+            let wasmSrs = conversion[f].pointsToRust(mlSrs);
+
+            srs = setSrs(wasmSrs);
+          } catch (e) {
```
dannywillems, line 110: should it not be around cache.read then? We must be sure it comes from the cache.read function. Many other exceptions can be raised.

crypto/bindings/srs.ts
```diff
@@ -0,0 +1,215 @@
...
+                cache.write(header, bytes);
+              } catch (e) {}
+            }
+          }
+        }
+
+        srsStore[f][size] = srs;
+      }
+
+      // should we do call freeOnFinalize() and expose a function to clean the SRS cache?
```
dannywillems, line 131: Maybe add a TODO or link an issue to keep track of this.

Trivo25 (approved):
lgtm, cant really say too much about the Rust part tho!

kimchi/wasm/src/srs.rs
dannywillems: Mmh, I see. I understand your point, and I do agree (I hate macros...). However, IMO, it is good to stay with one convention, and maybe initiate a refactorization after.
Not a blocking comment for this PR, only an opinion.

scripts/update-wasm-and-types.sh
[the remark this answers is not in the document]
dannywillems: It is used in dune.linker.inc to change the linker. I don't know if it is really useful here for JS compilation.
It seems to be used for Nix :grimacing: (search `linker` in nix/ocaml.nix). Ah, Nix users...

dannywillems (approved):

crypto/bindings/srs.ts
mitschabaude: hm, but `read()` can be very simple. it just returns bytes contained in a file. what if the file is garbage? then the decoding here will fail, not `read()`. I think we shouldn't trust the file system to return something that can be deserialized. but we also shouldn't treat a failed cache read as a hard error that crashes the process, since we can just regenerate the data and write back a correct version.

maybe to address your point, the conversion logic should be tested as a unit. and there should be an easy way to test a `Cache`. maybe we can just introduce a "caches may fail" mode which is used when debugging your cache, where errors aren't swallowed

dannywillems: Something explicit in case of an error would be nice to have. It will help debugging in this case, even not very often happening.

crypto/bindings/srs.ts
mitschabaude: I wrote this before the cache had a `canWrite` property. so I had to swallow the error here to allow dummy / non-writing caches without spamming the console. However, now that there is `canWrite`, I will actually log the error here

crypto/bindings/srs.ts
mitschabaude: I'm addressing this by adding a `debug` parameter to the `Cache`. Also I'm abstracting away the try-catch acrobatics into helpers

PR #194, opened 2023-10-26
Implement NOT gadget

ymekuria:
# Description
This PR adds a not function for finite fields in the case of constants.

A bitwise `NOT` is an operation that returns `1` in each bit position if the corresponding bit of the operand is `0`, and returns `0` if the corresponding bit of the operand is `1`.

# Tested
This work is tested in the upper layers of o1js with unit and e2e tests that are built on top of the exposed gate constraint.

🔗 o1js: https://github.com/o1-labs/o1js/pull/1198


crypto/finite_field.ts
```diff
@@ -122,7 +122,9 @@ function createField(p: bigint, t: bigint, twoadicRoot: bigint) {
     sizeInBits: 255,
     t,
     twoadicRoot,
-
+    not(x: bigint) {
+      return ~x
```
Trivo25, line 126: hm I think the gadget actually works slightly different than the JavaScript operation `~` (hence the tests fail) - so we have to wrap it to get valid field elements. The gadget operates over only over `bits`, whereas negating a `bigint` and taking that as the new Field element operates over the entire modulus of the field, so we have to simply limit this operations to not exceed `bits`

```ts
let a = Provable.witness(Field, () => new Field(5n));
let c = Gadgets.not(a, 8);
Provable.log(c);
Provable.log(Field(~5)
// 250
// 28948022309329048855892746252171976963363056481941560715954676764349967630331 
```


we can either do that by using something like
```ts
const not = (x: bigint, bits: number) => 2n ** BigInt(bits) - (x + 1n);
```

or also calling into the XOR gadget in the constant case, which should do the same thing

mitschabaude (approved):

PR #201, opened 2023-11-01
Expose remaining gates

mitschabaude:
* expose all gates
* make gate definitions shorter
* put gates in the same order as in `plonk_constraint_system.ml`


ocaml/lib/snarky_bindings.ml
```diff
@@ -113,43 +115,218 @@ module Field' = struct
...
+
   let generic sl l sr r so o sm sc =
-    Impl.with_label "generic_gate" (fun () ->
-        Impl.assert_
-          { annotation = Some __LOC__
-          ; basic =
-              Kimchi_backend_common.Plonk_constraint_system.Plonk_constraint.T
-                (Basic { l = (sl, l); r = (sr, r); o = (so, o); m = sm; c = sc })
-          } )
+    add_gate "generic_gate"
```
Trivo25, line 133: nit: I think `generic` is good enough in this context, no need to explicitly specify `_gate`

ocaml/lib/snarky_bindings.ml
```diff
@@ -113,43 +115,218 @@ module Field' = struct
...
+      (Basic { l = (sl, l); r = (sr, r); o = (so, o); m = sm; c = sc })
+
+  let poseidon state = add_gate "poseidon" (Poseidon { state })
+
+  let ec_add p1 p2 p3 inf same_x slope inf_z x21_inv =
+    add_gate "ec_add"
+      (EC_add_complete { p1; p2; p3; inf; same_x; slope; inf_z; x21_inv }) ;
+    (* TODO: do we need this? *)
```
Trivo25, line 141: do you mean the return value?

Trivo25 (approved):
this makes the code much more cleaner, really like that!

ocaml/lib/snarky_bindings.ml
mitschabaude: yeah

Trivo25: I think we dont. That was probably my fault when I originally rewrote the `Group.add` function in TypeScript

https://github.com/o1-labs/o1js/blob/main/src/lib/group.ts#L173-L181

can probably just return

```ts
(x3, y3)
```

PR #203, opened 2023-11-02
Foreign field backend

mitschabaude:
This PR generalizes the bigint finite field code to work for all fields.
This enables foreign field tests in https://github.com/o1-labs/o1js-bindings/pull/203

We also add a file with a range of concrete finite fields that can be used for testing and in examples.

### DIFF
--- a/crypto/finite_field.ts
+++ b/crypto/finite_field.ts
@@ -1,7 +1,7 @@
 import { bytesToBigInt } from './bigint-helpers.js';
 import { randomBytes } from './random.js';
 
-export { Fp, Fq, FiniteField, p, q, mod, inverse };
+export { createField, Fp, Fq, FiniteField, p, q, mod, inverse };
 
 // CONSTANTS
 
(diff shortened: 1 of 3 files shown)


mrmr1993 (approved):

crypto/finite_field.ts
```diff
@@ -206,3 +225,26 @@ function createField(p: bigint, t: bigint, twoadicRoot: bigint) {
     },
   };
 }
+
+/**
+ * Compute constants to instantiate a finite field just from the modulus
+ */
+function computeFieldConstants(p: bigint) {
+  // figure out the factorization p - 1 = 2^M * t
+  let oddFactor = p - 1n;
```
rbonichon, line 234: Nit: the names make it somewhat hard to read.

 Is `p - 1n` necessarily odd? If `p` is a prime number, as its naming can also suggest, then `oddFactor` usually starts by being even (and stays so by subsequent ` >> 1n`). In any case. it's not ensured by the parameter type.

Maybe that's the right name, though. Can you refer to a source where this is used/defined?

rbonichon (approved):
Only minor comments

crypto/finite_field.ts
mitschabaude: The code right below that line divides `oddFactor` by 2 as many times as needed to become odd.

so this code, as the comment hints at (but could've used clearer naming), figures out the factorization
`p - 1 = 2^twoadicity * oddFactor`, for a prime `p` (well for any number)

mitschabaude: we need the `oddFactor` in some algorithms -- for example, right below in the same function to find a primitive root of unity

PR #208, opened 2023-11-17
Lookup tests assertions

volhovm:
Original PR in proof-systems: https://github.com/o1-labs/proof-systems/pull/1263
Compatibility PR in mina: https://github.com/MinaProtocol/mina/pull/14549


dannywillems (approved):
Annoying, but I think you need to run `cargo fmt`. The line seems to long. Approving but would be nice to fix it please :).

PR #209, opened 2023-11-17
Make ECDSA more efficient with GLV

mitschabaude:
Companion of https://github.com/o1-labs/o1js/pull/1257

* Given a curve endomorphism with constants `endoScalar`, `endoBase`, implements methods to decompose a scalar, and to do faster scalar multiplication using that decomposition ("GLV method")
* Also adds a method to automatically compute `endoScalar` and `endoBase` when they exist and are not already given as parameters (i.e. for secp256k1)
* We can statically bound the max number of bits that decomposed scalars have. For Pallas and Vesta, we get 127 bits while for secp256k1 we get 128 bits. In both cases, this reduces the scalar bits that we need to process in the circuit exactly by half (while doubling the number of points to scale)


crypto/elliptic-curve-endomorphism.ts
```diff
@@ -0,0 +1,300 @@
...
+        glvData
+      );
+      return projectiveToAffine(sGProj, Field.modulus);
+    },
+  };
+}
+
+/**
+ * GLV decomposition, named after the authors Gallant, Lambert and Vanstone who introduced it:
+ * https://iacr.org/archive/crypto2001/21390189.pdf
```
mitschabaude, line 83: this is where the magic happens!

crypto/elliptic-curve-endomorphism.ts
```diff
@@ -0,0 +1,300 @@
...
+function computeGlvData(q: bigint, lambda: bigint) {
+  let [[v00, v01], [v10, v11]] = egcdStopEarly(lambda, q);
+  let det = v00 * v11 - v10 * v01;
+
+  // upper bounds for
+  // |s0| <= 0.5 * (|v00| + |v01|)
+  // |s1| <= 0.5 * (|v10| + |v11|)
+  let maxS0 = ((abs(v00) + abs(v01)) >> 1n) + 1n;
+  let maxS1 = ((abs(v10) + abs(v11)) >> 1n) + 1n;
+  let maxBits = log2(max(maxS0, maxS1));
```
mitschabaude, line 242: it's essential that we are able to statically compute the maximum number of bits (`maxBits`) that scalars have after GLV decomposition, with a reliable upper bound. Only with this knowledge way can we leverage the gains of reduced bit lengths in a static circuit.

Trivo25 (approved):

PR #210, opened 2023-11-21
make `Fp.rot` take variable bit sizes 

Trivo25:
Preparation for rotation over 32 bits

### DIFF
--- a/crypto/finite_field.ts
+++ b/crypto/finite_field.ts
@@ -201,25 +201,15 @@ function createField(
     },
     rot(
       x: bigint,
-      bits: number,
+      bits: bigint,
       direction: 'left' | 'right' = 'left',
-      maxBits = 64
+      maxBits = 64n
     ) {
-      let bitArray = x.toString(2).split('').reverse().map(Number);
-      let binary: number[] =
-        bitArray.length >= maxBits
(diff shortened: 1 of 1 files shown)


crypto/finite_field.ts
```diff
@@ -199,12 +199,17 @@ function createField(
     fromBigint(x: bigint) {
       return mod(x, p);
     },
-    rot(x: bigint, bits: number, direction: 'left' | 'right' = 'left') {
+    rot(
+      x: bigint,
+      bits: number,
+      direction: 'left' | 'right' = 'left',
+      maxBits = 64
+    ) {
```
mitschabaude, line 207: maybe we can use native bit shifts to make the code faster and cleaner?

```ts
    rot(
      x: bigint,
      bits: bigint,
      direction: 'left' | 'right' = 'left',
      maxBits = 64n
    ) {
      if (direction === 'right') bits = maxBits - bits;
      let full = x << bits;
      let excess = full >> maxBits;
      let shifted = full & ((1n << maxBits) - 1n);
      return shifted | excess;
    } 
```

Trivo25: yep, good idea!

mitschabaude (approved):

PR #224, opened 2023-12-18
Fix stack overflow for large witnesses

mitschabaude:
companion for https://github.com/o1-labs/o1js/pull/1334

### DIFF
--- a/ocaml/lib/snarky_bindings.ml
+++ b/ocaml/lib/snarky_bindings.ml
@@ -11,16 +11,18 @@ type field = Impl.field
 
 (* light-weight wrapper around snarky-ml core *)
 
-let typ (size_in_fields : int) : (Field.t array, field array) Typ.t =
-  Typ
-    { var_to_fields = (fun fields -> (fields, ()))
-    ; var_of_fields = (fun (fields, _) -> fields)
-    ; value_to_fields = (fun fields -> (fields, ()))
-    ; value_of_fields = (fun (fields, _) -> fields)
-    ; size_in_field_elements = size_in_fields
-    ; constraint_system_auxiliary = (fun _ -> ())
-    ; check = (fun _ -> Impl.Internal_Basic.Checked.return ())
-    }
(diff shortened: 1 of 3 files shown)


ocaml/overrides.js
```diff
@@ -1,6 +1,17 @@
...
+//Provides: caml_wrap_exception const (const)
+//Requires: caml_global_data,caml_string_of_jsstring,caml_named_value
+//Requires: caml_return_exn_constant
+function caml_wrap_exception(e) {
+  if (e instanceof Array) return e;
+  if (e instanceof globalThis.Error && caml_named_value('jsError'))
+    return [0, caml_named_value('jsError'), e];
+  //fallback: wrapped in Failure
+  return [0, caml_global_data.Failure, caml_string_of_jsstring(String(e))];
+}
```
mitschabaude, line 13: this fixes that stack overflows used to be thrown as an array. it overrides the `caml_wrap_exception` function from JSOO, which has a special case for stack overflows, which doesn't presever the stack trace.

this simply removes the special case, and thus stack overflows end up in the first `if` case, where we preserve the original error `e` which can be later be re-thrown in JS.

ocaml/lib/snarky_bindings.ml
```diff
@@ -11,7 +11,18 @@ type field = Impl.field
...
+  ; var_of_fields = (fun (fields, _) -> fields)
+  ; value_to_fields = (fun fields -> (fields, ()))
+  ; value_of_fields = (fun (fields, _) -> fields)
+  ; size_in_field_elements = 0
+  ; constraint_system_auxiliary = (fun _ -> ())
+  ; check = (fun _ -> Impl.Internal_Basic.Checked.return ())
+  }
+
+let typ (size_in_field_elements : int) : (Field.t array, field array) Typ.t =
+  Typ { empty_typ with size_in_field_elements }
```
mitschabaude, line 25: the original implementation using `Typ.array` was causing stack overflows in (at least) the `check` function, because it operated on an OCaml list with a recursive function

Trivo25 (approved):

PR #232, opened 2024-01-08
[rampup] Add support to build bindings using mina as a submodule in o1js repo

MartinMinkov:
o1js: https://github.com/o1-labs/o1js/pull/1358


crypto/bindings.ts
mitschabaude, line 1: same question as on the berkeley PR, why do we add this file?

MartinMinkov: Good catch, resolved!

mitschabaude (approved):

PR #251, opened 2024-02-22
Improve performance by reverting Apple silicon workaround (root problem is fixed upstream)

jackryanservia:
Bindings PR for o1-labs/o1js#1456


js/node/node-backend.js
```diff
@@ -82,7 +83,7 @@ async function withThreadPool(run) {
 async function initThreadPool() {
   if (!isMainThread) return;
   workersReady = new Promise((resolve) => (workersReadyResolve = resolve));
-  await wasm.initThreadPool(getEfficientNumWorkers(), filename);
+  await wasm.initThreadPool(workers.numWorkers ?? os.cpus().length, filename);
```
mitschabaude, line 86: Instead of `os.cpus().length`, a better default could be `os.availableParallelism() - 1` https://nodejs.org/api/os.html#osavailableparallelism

I thought the `-1` makes sense because these workers are started in addition to the main thread 🤔

js/web/web-backend.js
```diff
@@ -55,7 +55,7 @@ async function withThreadPool(run) {
   if (workerPromise === undefined)
     throw Error('need to initialize worker first');
   let worker = await workerPromise;
-  numWorkers ??= await getEfficientNumWorkers();
+  numWorkers ??= workers.numWorkers ?? navigator.hardwareConcurrency;
```
mitschabaude, line 58: here as well, `navigator.hardwareConcurrency - 1` would make sense to me personally, although it probably won't make much of a difference (on the intel CPUs that I have, benefits from parallelism steeply diminish after the 4th thread or so)

mitschabaude (approved):
approved with a suggestion

js/node/node-backend.js
jackryanservia: Ahhhhh! In that case, I will switch it back to `-1`. I didn't understand that the first thread also did work so I figured it would have extra time for a worker. (I only added `-1` because it ended up being faster in practice for me.) But if I add `-1` back in, it should still work perfectly for a machine with only a single core?

mitschabaude: > But if I add -1 back in, it should still work perfectly for a machine with only a single core?

ah, great question lol. probably not. maybe `Math.max(n - 1, 1)`?

PR #268, opened 2024-04-15
Remove top-level await

mitschabaude:
companion for https://github.com/o1-labs/o1js/pull/1583

### DIFF
--- a/js/web/web-backend.js
+++ b/js/web/web-backend.js
@@ -8,7 +8,7 @@ import {
 import o1jsWebSrc from 'string:../../../web_bindings/o1js_web.bc.js';
 import { workers } from '../../../lib/proof-system/workers.js';
 
-export { initO1, withThreadPool };
+export { initializeBindings, withThreadPool };
 
 let wasm = plonkWasm();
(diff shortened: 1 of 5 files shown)


js/web/web-backend.js
```diff
@@ -23,13 +21,11 @@ let workerPromise;
  */
 let numWorkers = undefined;
 
-let isInitialized = false;
+async function initializeBindings() {
+  wasm = plonkWasm();
```
mitschabaude, line 25: I moved this here to make the initial loading time even faster (< 20ms)

ymekuria (approved):

MartinMinkov (approved):

js/web/web-backend.js
MartinMinkov: Is there any use in returning from the function early if `wasm` is already defined? Let's say we accidentally call `initializeBindings()` twice, does that mean we have to wait for loading twice?

mitschabaude: Yeah that's very important! I moved the early return logic into snarky.web.js, through which this function will always be called

PR #277, opened 2024-06-17
Optional feature flags in `Pickles.sideLoaded`

Trivo25:


ocaml/lib/pickles_bindings.ml
```diff
@@ -77,6 +77,16 @@ type pickles_rule_js =
...
 
+let map_feature_flags_option
+    (feature_flags_ : bool option Pickles_types.Plonk_types.Features.t) =
+  Pickles_types.Plonk_types.Features.map feature_flags_ ~f:(function
+    | Some true ->
+        Pickles_types.Opt.Flag.Yes
+    | Some false ->
+        Pickles_types.Opt.Flag.No
+    | None ->
+        Pickles_types.Opt.Flag.Maybe )
```
mitschabaude, line 88: nicely done!
btw if you'd want to model `Pickles_types.Opt.Flag.t` in TS, as an enum without contents it maps to `0 | 1 | 2`, where the order is determined by the order that `Yes | No | Maybe` appear in the enum definition

mitschabaude (approved):

ocaml/lib/pickles_bindings.ml
```diff
@@ -298,9 +308,14 @@ module Choices = struct
           |> Promise_js_helpers.of_js
           |> Promise.map ~f:(finish_circuit prevs self)
         in
-
         { identifier = Js.to_string rule##.identifier
-        ; feature_flags = rule##.featureFlags
+        ; feature_flags =
```
Trivo25, line 312: I changed the type of the pickles rule here as well to match the rest (less different types we have to consinder in javascript or here)

PR #278, opened 2024-06-27
Add Detailed Error Handling for Invalid Struct Type Usage

MartinMinkov:
## Summary
This PR addresses an issue with the check() method in circuit classes extending Struct. 

## Changes
- Add a runtime check for `check` which checks if we are running on a `Struct` and throw an error if so

## Test Case

```ts
import { Struct, Bool, Field } from 'o1js';

class WrappedStruct extends Struct({
  wrappedA: Struct,
}) {}

class WrappedBool extends Struct({
  wrappedB: Bool,
}) {}

class WrappedStructGood extends Struct({
  wrappedA: WrappedBool,
}) {}

let notBool = new Bool(false);
notBool.value = new Field(2n).value; // set to a non-bool

let wrappedBool = new WrappedBool({
  wrappedB: notBool,
});
let wrappedWrappedBool = new WrappedStruct({
  wrappedA: wrappedBool,
});

let wrappedWrappedBoolGood = new WrappedStructGood({
  wrappedA: wrappedBool,
});

WrappedBool.check(wrappedBool); // should throw error correctly from check (invalid Bool)
WrappedStructGood.check(wrappedWrappedBoolGood); // should throw error correctly from check (invalid Bool)
WrappedStruct.check(wrappedWrappedBool); // should error here with thrown error (cannot use Struct here)
```

### DIFF
--- a/lib/provable-generic.ts
+++ b/lib/provable-generic.ts
@@ -178,13 +178,6 @@ function createDerivers<Field>(): {
 
       if (isProvable(typeObj)) return typeObj.check(obj);
 
-      if (typeof typeObj === 'function') {
-        throw new Error(
-          `provable: invalid type detected. Functions are not supported as types. ` +
-            `Ensure you are passing an instance of a supported type or an anonymous object.\n`
-        );
-      }
-
       if (display(typeObj) === 'Struct') {
         throw new Error(
           `provable: cannot run check() on 'Struct' type. ` +
@@ -198,6 +191,13 @@ function createDerivers<Field>(): {
             `class MyStruct extends Struct({\n` +
             `  fieldA: MySpecificStruct, // Use the specific struct type\n` +
             `}) {}\n`
+        );
+      }
+
+      if (typeof typeObj === 'function') {
+        throw new Error(
+          `provable: invalid type detected. Functions are not supported as types. ` +
+            `Ensure you are passing an instance of a supported type or an anonymous object.\n`
         );
       }
 
--- a/lib/provable-generic.ts
+++ b/lib/provable-generic.ts
@@ -173,6 +173,21 @@ function createDerivers<Field>(): {
       if (!complexTypes.has(typeof typeObj))
         throw Error(`provable: unsupported type "${typeObj}"`);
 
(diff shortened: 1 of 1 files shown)


lib/provable-generic.ts
```diff
@@ -173,6 +173,21 @@ function createDerivers<Field>(): {
       if (!complexTypes.has(typeof typeObj))
         throw Error(`provable: unsupported type "${typeObj}"`);
 
+      if (display(typeObj) === 'Struct')
```
mitschabaude, line 176: * I think this should be moved below the `if (isProvable(...)) { ... }` branch so we don't prevent valid provable types called `Struct`
* To further harden what types we accept, I think we should add another check with generalizes this one: If, after the `isProvable()` case, we have `typeof typeObj === "function"`, then this is almost definitely a mistake and we should throw an error (a more generic error than this one of course, so keep this error for the specific case of Struct). the final case where we recursive into the object properties should only be done for anonymous objects.

MartinMinkov: Added the extra check! If the error message isn't good enough, happy to change it!

lib/provable-generic.ts
```diff
@@ -178,6 +178,30 @@ function createDerivers<Field>(): {
 
       if (isProvable(typeObj)) return typeObj.check(obj);
 
+      if (typeof typeObj === 'function') {
+        throw new Error(
+          `provable: invalid type detected. Functions are not supported as types. ` +
+            `Ensure you are passing an instance of a supported type or an anonymous object.\n`
+        );
+      }
```
mitschabaude, line 186: Struct is a function though, so this has to come after the detailed error about Struct!

lib/provable-generic.ts
mitschabaude: Thanks!

lib/provable-generic.ts
MartinMinkov: [Resolved!](https://github.com/o1-labs/o1js-bindings/pull/278/commits/cbdbff3d9c317aa1131667cd56f2a8a374640436)

mitschabaude (approved):

PR #281, opened 2024-07-13
Update AffineCurve equal method to account for points at infinity

ymekuria:
## Summary

This PR addresses an issue with the the `equal()` method in `AffineCurve`.  The method ignores the possibility of either point in the comparison being a point at infinity. 

## Impact

This issue could lead to equality checks that may succeed when they should fail, or vice versa, when the point at infinity is involved.

## Changes

- New types `PointAtInfinity` and `FinitePoint` were added.
- The type definition of `GroupAffine` was updated to be a union type of the `PointAtInfinity`  and `FinitePoint`.
- The `equal` method in `CreateCurveAffine`  was implemented to correctly handle the cases when either point is at infinity. 

o1js: https://github.com/o1-labs/o1js/pull/1758


Trivo25:
This change causes a bubble-up effect in o1js as well, can you address this in a parent PR into o1js? If this introduces any breaking changes, we should consider targeting V2 if we cannot avoid breaking the API

ymekuria:
> This change causes a bubble-up effect in o1js as well, can you address this in a parent PR into o1js? If this introduces any breaking changes, we should consider targeting V2 if we cannot avoid breaking the API

I'll link the parent PR in this PR.

crypto/elliptic-curve.ts
```diff
@@ -58,7 +58,9 @@ const a = 0n;
 const projectiveZero = { x: 1n, y: 1n, z: 0n };
 
 type GroupProjective = { x: bigint; y: bigint; z: bigint };
-type GroupAffine = { x: bigint; y: bigint; infinity: boolean };
+type PointAtInfinity = { x: bigint; y: bigint; infinity: true };
+type FinitePoint = { x: bigint; y: bigint; infinity: false };
+type GroupAffine = PointAtInfinity | FinitePoint;
```
mitschabaude, line 63: isn't this the same type as before now? since `boolean = true | false`?

I thought the version where you had `type PointAtInfinity = { infinity: true };` made sense, but not sure how cumbersome it would be to use. If a lot of work, I'm also fine with keeping the type

ymekuria: > I thought the version where you had type PointAtInfinity = { infinity: true }; made sense, but not sure how cumbersome it would be to use. If a lot of work, I'm also fine with keeping the type

This `type PointAtInfinity = { infinity: true };` was my original approach and made sense to me also. When I added this type, changes were necessary in many methods that used the `GroupAffine` type in the `eliptic-curve.ts` and `elliptic-curve-endomorphism.ts` files because those methods were destructuring `x,y, infinity` in the parameters.

I can easily revert to my previous approach and make the necessary changes in the codebase. I only refactored because it required the least amount of code changes.

mitschabaude: Ok, I think going with the least amount of changes is probably best!

mitschabaude (approved):

PR #280, opened 2024-07-11
[v2] Fix MayUseToken vulnerability in AccountUpdates

MartinMinkov:
o1js: https://github.com/o1-labs/o1js/pull/1741
mina: https://github.com/MinaProtocol/mina/pull/15808

PR #283, opened 2024-07-15
[Main] Fix MayUseToken vulnerability in AccountUpdates

MartinMinkov:
Adds a custom JS deriver for `MayUseToken` without any overrides. Since this is landing in `main`, and we can't break compatibility yet, the `check` function is not overridden. This PR just provides the machinery to handle the new deriving logic. The `check` function will be overridden in a [different PR](https://github.com/o1-labs/o1js-bindings/pull/280), targeting v2

o1js: https://github.com/o1-labs/o1js/pull/1750
mina: https://github.com/MinaProtocol/mina/pull/15808


mina-transaction/derived-leaves.ts
```diff
@@ -195,6 +199,20 @@ function createAuthRequired<
...
+};
+function createMayUseToken<
+  Field,
+  Bool,
+  Base extends GenericSignable<MayUseToken<Bool>, MayUseToken<boolean>, Field>
+>(base: Base, _Bool: GenericSignableBool<Field, Bool>) {
+  return {
+    ...base,
+  };
+}
```
mitschabaude, line 214: since `createMayUseToken()` is now a no-op (and stays that way in `v2`) you could remove it!

mitschabaude (approved):

mina-transaction/derived-leaves.ts
MartinMinkov: Are you sure? I'm not sure what you mean by it staying that way in v2, since we still have it [here](https://github.com/o1-labs/o1js-bindings/pull/280/files#diff-ba61e55f291e2aba4ef07e38b755b4160072dac8e62ef8901574d22a8bf21d33R202). I might be mistaken but I think this is still needed to get the bindings to build properly. Otherwise, I get errors from the generated files, e.g.

```sh
src/bindings/mina-transaction/gen/transaction-bigint.ts:113:3 - error TS2693: 'MayUseToken' only refers to a type, but is being used as a value here.

113   MayUseToken,
      ~~~~~~~~~~~

src/lib/testing/random.ts:117:3 - error TS2693: 'MayUseToken' only refers to a type, but is being used as a value here.

117   MayUseToken.fromJSON
      ~~~~~~~~~~~

src/lib/testing/random.ts:164:3 - error TS2322: Type 'Random<unknown>' is not assignable to type 'Random<{ parentsOwnToken: boolean; inheritFromParent: boolean; }>'.
  Type 'unknown' is not assignable to type '{ parentsOwnToken: boolean; inheritFromParent: boolean; }'.

164   MayUseToken: mayUseToken,
      ~~~~~~~~~~~

  src/bindings/mina-transaction/gen/transaction-bigint.ts:92:3
    92   MayUseToken: Signable<
         ~~~~~~~~~~~
    The expected type comes from property 'MayUseToken' which is declared here on type 'Generators'

src/lib/testing/random.ts:273:44 - error TS2693: 'MayUseToken' only refers to a type, but is being used as a value here.

273   MayUseToken: mapWithInvalid(mayUseToken, MayUseToken.toJSON),
```

mitschabaude: I meant that this function only takes 'base' and copies it to a new object. Same in the v2 PR. Of course you still need to define 'MayUseToken' above, but you don't need this function for it

PR #284, opened 2024-07-16
Accept .provable and some other type improvements

mitschabaude:
companion for https://github.com/o1-labs/o1js/pull/1751

handles `.provable` in type deriver methods & inferring types


lib/provable-generic.ts
```diff
@@ -461,28 +513,35 @@ type InferPrimitiveJson<P extends Primitive> = P extends typeof String
...
 type NestedProvable<Field> =
   | Primitive
+  | { provable: GenericProvable<any, any, Field> }
   | GenericProvable<any, any, Field>
   | [NestedProvable<Field>, ...NestedProvable<Field>[]]
   | NestedProvable<Field>[]
   | { [key: string]: NestedProvable<Field> };
 
-type InferProvable<A, Field> = A extends Constructor<infer U>
+type InferProvable<A, Field> = A extends { provable: Constructor<infer U> }
```
Trivo25, line 526: nothing better than a nested conditional type :D imagine if TS would allow `if`, `else` and brackets here

Trivo25 (approved):

PR #285, opened 2024-07-16
Update array handling in Provable.toFields 

ymekuria:
## Summary

This PR addresses an issue with incorrect array handling in the `provable.toFields()` method. 

## Impact

The resulting conversion to fields could be invalid. Certain field elements may be silently ignored.

## Changes

- Dynamic checks added in `toFields` to make sure that `typeObj` and `obj` have the same length.
- Check added to ensure `obj` is an array if `typeObj` is an array.
- Error message added if `typeObj` and `obj` have different lengths.
- Error message added if `typeObj` is an array and `obj` is not.

o1js: https://github.com/o1-labs/o1js/pull/1791


lib/provable-generic.ts
```diff
@@ -96,8 +96,17 @@ function createDerivers<Field>(): {
       if (!complexTypes.has(typeof typeObj))
         throw Error(`provable: unsupported type "${typeObj}"`);
 
-      if (Array.isArray(typeObj))
+      if (Array.isArray(typeObj)) {
+        if (!Array.isArray(obj)) {
+          throw new Error(`Expected an array for type, but got ${typeof obj}`);
```
MartinMinkov, line 101: Can you change this to `throw Error`? We have a convention to use the latter.

mitschabaude (approved):

PR #286, opened 2024-07-17
Canonical representation of provable types

mitschabaude:
companion of https://github.com/o1-labs/o1js/pull/1759


lib/from-layout.ts
```diff
@@ -349,6 +349,13 @@ function ProvableFromLayout<
       check(value: T): void {
         check(typeData, value);
       },
+      // TODO implement properly
```
ymekuria, line 352: Thanks for adding a reminder in the comments. Do you mind also creating an issue for this so it doesn't get lost?

ymekuria (approved):
Just a minor comment.

PR #287, opened 2024-07-22
feat(finite-field): add canonicalize function 

MartinMinkov:


crypto/finite-field.ts
```diff
@@ -266,89 +266,112 @@ function createField(
...
     },
     fromNumber(x: number) {
-      return mod(BigInt(x), p);
+      return canonicalize(BigInt(x));
     },
     fromBigint(x: bigint) {
-      return mod(x, p);
+      return canonicalize(x);
     },
     rot(
```
mitschabaude, line 355: `rot()`, `leftShift()` and `rightShift()` are not finite field operations, and reducing modulo the field size makes no sense here -- so please remove `canonicalize()` in inputs in these functions (they shouldn't have been put into `finite-field.ts` in the first place)

crypto/finite-field.ts
```diff
@@ -266,89 +266,112 @@ function createField(
...
+        p,
+        oddFactor,
+        twoadicRoot,
+        twoadicity
+      );
+      return result !== undefined ? canonicalize(result) : undefined;
     },
     power(x: bigint, n: bigint) {
-      return power(x, n, p);
+      return canonicalize(power(x, n, p));
```
mitschabaude, line 330: power already returns a reduced result, can you remove `canonicalize()` here?

crypto/finite-field.ts
```diff
@@ -266,89 +266,112 @@ function createField(
...
-      return mod(x * yinv, p);
+      return canonicalize(x * yinv);
     },
     square(x: bigint) {
-      return mod(x * x, p);
+      return canonicalize(x * x);
     },
     isSquare(x: bigint) {
-      return isSquare(x, p);
+      return isSquare(canonicalize(x), p);
```
mitschabaude, line 317: this step should be part of the `isSquare()` function -- can you do `mod(., p)` on the input there and remove `canonicalize()` here?

crypto/finite-field.ts
```diff
@@ -266,89 +266,112 @@ function createField(
...
     },
     sqrt(x: bigint) {
-      return sqrt(x, p, oddFactor, twoadicRoot, twoadicity);
+      const result = sqrt(
+        canonicalize(x),
```
mitschabaude, line 321: this step seems to fix a bug and so is an important part of the `sqrt()` function -- can you do `mod(., p)` on the input there and remove `canonicalize()` here?

crypto/finite-field.ts
```diff
@@ -266,89 +266,112 @@ function createField(
...
+        p,
+        oddFactor,
+        twoadicRoot,
+        twoadicity
+      );
+      return result !== undefined ? canonicalize(result) : undefined;
```
mitschabaude, line 327: the result returned by `sqrt()` is already reduced so please remove `canonicalize()` here

(i.e. you can `return sqrt(...)` directly)

crypto/finite-field.ts
```diff
@@ -266,89 +266,112 @@ function createField(
...
     power(x: bigint, n: bigint) {
-      return power(x, n, p);
+      return canonicalize(power(x, n, p));
     },
     dot(x: bigint[], y: bigint[]) {
       let z = 0n;
       let n = x.length;
       for (let i = 0; i < n; i++) {
-        z += x[i] * y[i];
+        z += canonicalize(x[i]) * canonicalize(y[i]);
```
mitschabaude, line 336: this is not needed, it's handled by the final call to `mod()` / `canonicalize()` before returning

(in fact, I introduced the `dot()` method precisely to save calls to `mod()` compared to just using multiplication and addition, for performance of Poseidon)

crypto/finite-field.ts
```diff
@@ -266,89 +266,112 @@ function createField(
...
       }
-      return mod(z, p);
+      return canonicalize(z);
     },
     equal(x: bigint, y: bigint) {
-      return mod(x - y, p) === 0n;
+      return canonicalize(x) === canonicalize(y);
```
mitschabaude, line 341: at first glance I thought this is less efficient than the original, but since inputs are typically canonical, it should actually be more efficient since we never have to `mod()`!

crypto/finite-field.ts
```diff
@@ -266,89 +266,112 @@ function createField(
...
     },
     isEven(x: bigint) {
-      return !(x & 1n);
+      return !(canonicalize(x) & 1n);
     },
     random() {
-      return randomField(p, sizeInBytes, hiBitMask);
+      return canonicalize(randomField(p, sizeInBytes, hiBitMask));
```
mitschabaude, line 347: `randomField()` already returns a canonical result, so please remove `canonicalize()` here

crypto/finite-field.ts
```diff
@@ -266,89 +266,112 @@ function createField(
...
+   * Ensures the given bigint is in its canonical form within the field.
+   * This function is added to address issues raised in a security audit,
+   * where certain operations assumed inputs were already in canonical form.
+   *
+   * It helps prevent incorrect results and infinite loops that could occur
+   * when non-canonical representations are used in operations like {@link isSquare()},
+   * {@link sqrt()}, and others. By using this function, we ensure consistent behavior
+   * across all field operations regardless of the input's initial representation.
+   */
+  function canonicalize(x: bigint) {
```
Trivo25, line 279: I actually don't see how this is different to `mod(x, p)` and why we should replace`mod` with this new function - we could simply check inputs (and outputs) by calling `mod(x, p)` to assert that they are canonical, like this:

```diff
- return sqrt(x, p, oddFactor, twoadicRoot, twoadicity);
+ return sqrt(mod(x, p), p, oddFactor, twoadicRoot, twoadicity);
```

mitschabaude (commented):
some comments!

comment on performance: `canonicalize()` is more efficient in the case its input is already canonical (in that case, it avoids the `mod()` call thanks to the inequality checks)
it's less efficient in the case the input needs to be reduced. 

overall, I think this trade-off is worth it, so I support changing `canonicalize()` to `mod()` all over the field module even though they do the same!

crypto/finite-field.ts
MartinMinkov: Fixed! https://github.com/o1-labs/o1js-bindings/pull/287/commits/145a9286ab198a2156967bb8bdca6a45273827fa

crypto/finite-field.ts
MartinMinkov: Fixed!

crypto/finite-field.ts
MartinMinkov: Fixed! https://github.com/o1-labs/o1js-bindings/pull/287/commits/145a9286ab198a2156967bb8bdca6a45273827fa

crypto/finite-field.ts
MartinMinkov: Fixed! https://github.com/o1-labs/o1js-bindings/pull/287/commits/145a9286ab198a2156967bb8bdca6a45273827fa

crypto/finite-field.ts
MartinMinkov: Fixed!

crypto/finite-field.ts
MartinMinkov: Removed the `canonical` function and just replaced with inline checks to mod here! https://github.com/o1-labs/o1js-bindings/pull/287/commits/145a9286ab198a2156967bb8bdca6a45273827fa

crypto/finite-field.ts
MartinMinkov: Fixed!

crypto/finite-field.ts
MartinMinkov: My original thinking was that by introducing checks in certain methods and not others, it could lead to a similar situation in the future. By blanket applying a function to alter the output to all of these methods, we'd have a way similar way for all methods to move forward on. This turns out not to be a favourable choice though! 

I have reverted the changes and just applied `mod` to the inputs of the offending functions that cause the bugs outlined in the script above.

crypto/finite-field.ts
```diff
@@ -300,10 +300,10 @@ function createField(
       return mod(x * x, p);
     },
     isSquare(x: bigint) {
-      return isSquare(x, p);
+      return isSquare(mod(x, p), p);
     },
     sqrt(x: bigint) {
-      return sqrt(x, p, oddFactor, twoadicRoot, twoadicity);
+      return sqrt(mod(x, p), p, oddFactor, twoadicRoot, twoadicity);
```
mitschabaude, line 306: so what I meant in my other review here is that I'd like those changes to be made to the original `sqrt` and `isSquare` functions defined at the _top-level_ of this file. just o that we don't have two versions of `sqrt()`, one with a bug and another that fixes the bug

mitschabaude: I know that currently the original `sqrt()` / `isSquare()` are not even exported directly, but other basic functions are, and if someone decides to export `sqrt()` directly in the future they probably won't think of fixing its bug

mitschabaude (changes requested):
the following two methods also need `mod()` on their inputs to always behave correctly:
* `negate()` (can just do `mod(-x, p)`)
* `isEven()`

crypto/finite-field.ts
```diff
@@ -317,7 +317,9 @@ function createField(
       return mod(z, p);
     },
     equal(x: bigint, y: bigint) {
-      return mod(x - y, p) === 0n;
+      let x_ = x >= 0n && x < p ? x : mod(x, p);
```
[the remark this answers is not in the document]
MartinMinkov: fixed here! https://github.com/o1-labs/o1js-bindings/pull/287/commits/71f2e698dadcdfc62c76a72248c0df71cfd39d4c

crypto/finite-field.ts
MartinMinkov: fixed here! https://github.com/o1-labs/o1js-bindings/pull/287/commits/cc3f63228f6caa025a39b761b1e38faa8c50e69e

mitschabaude (approved):
sweet!

PR #295, opened 2024-08-22
Changes to work better with nix

Geometer1729:
The o1js-bindings part of https://github.com/o1-labs/o1js/pull/1800


mitschabaude:
why does this move the ocaml code in a subdirectory called `/js`?

Geometer1729:
> why does this move the ocaml code in a subdirectory called `/js`?

I'm not certain, it looks like they are pretty trivial modules (3 lines total) just reexporting stuff AFAICT my best guess is there's a build reason they need to be in the same directory as the javascript but separating the rest of the ocaml is possible and seemed organizationally better.

mitschabaude:
> > why does this move the ocaml code in a subdirectory called `/js`?
> 
> I'm not certain, it looks like they are pretty trivial modules (3 lines total) just reexporting stuff AFAICT my best guess is there's a build reason they need to be in the same directory as the javascript but separating the rest of the ocaml is possible and seemed organizationally better.



> > why does this move the ocaml code in a subdirectory called `/js`?
> 
> I'm not certain, it looks like they are pretty trivial modules (3 lines total) just reexporting stuff AFAICT my best guess is there's a build reason they need to be in the same directory as the javascript but separating the rest of the ocaml is possible and seemed organizationally better.

ok as long as the build keeps working also without nix, moving these files into their separate folder is fine with me.

however, I think the name `js` for a folder containing ocaml code (that's part of a largely JS project) is very confusing.
@Geometer1729 can you rename it, maybe to `entrypoints` or `js_entrypoints` or `jsoo_exports` or whatever makes its content clearer?

Geometer1729:
That's fair, js is not a great name as it's more like an interface than a library in js.

Would you be willing to check that it works without nix? Part of my motivation for this was that I couldn't get it to build without nix in the first place.

mitschabaude:
> Would you be willing to check that it works without nix? Part of my motivation for this was that I couldn't get it to build without nix in the first place.

I tried it, get this error which I'm pretty sure we had a workaround for before this refactor
image

mitschabaude:
so weird, because there _is_ a `config.mlh` in the `o1js` root

Geometer1729:
I'm not sure what to make of that or how to reproduce/troubleshoot it, @Trivo25 , do you remember a workaround/refactor relating to this? Do you have the same issue building without nix?

Trivo25:
>I'm not sure what to make of that or how to reproduce/troubleshoot it, @Trivo25 , do you remember a workaround/refactor relating to this? Do you have the same issue building without nix?

Are you referring to the `src/config.mlh` issue?

Geometer1729:
> Are you referring to the `src/config.mlh` issue?

Yes, probably should have quote replied, my bad.

Trivo25:
Works for me now! Looks like we had to re-add our opam dep repo because of some stale versions 

```sh
opam repository add --yes --all --set-default o1-labs https://github.com/o1-labs/opam-repository.git
opam switch import opam.export
```

ocaml/lib/dune
```diff
@@ -37,6 +38,7 @@
   fields_derivers
   fields_derivers.zkapps
   genesis_constants
+  genesis_constants.compiled
```
Trivo25, line 41: Lets make sure this is in sync with the latest `compatible` commit

Trivo25 (approved):
looks good! I could successfully build the binding without nix too (see comment above)!

PR #298, opened 2024-08-23
Update bindings

Trivo25:


MINA_COMMIT
```diff
@@ -1,2 +1,2 @@
 The mina commit used to generate the backends for node and web is
-4660786e0c6548fa6d05ec80d69d508520a4c3d1
+336b2bcd9a7e5d4f900b2dc141ded76f0c49b395
```
svv232, line 2: should this be `0d2f6f55623bdf99a38b6b56fd5fb74c5615e5dd` instead? I was just checking the head from https://github.com/o1-labs/o1js/pull/1804/commits

Trivo25: This is actually the commit of the mina repository that was used to generate the bindings! It doesn't really do anything and it's just automatically generated every time we generate the bindings

svv232: ahh i see

ymekuria (approved):

PR #299, opened 2024-09-04
Update instructions on building mina-transaction/gen/

Geometer1729:
Mina doesn't seem to have `make snarkyjs` anymore, but these bindings are regenerated by `npm run build:update-bindings` 

Here's the call to dune:
https://github.com/o1-labs/o1js-bindings/blob/ff3588199eaf8f423c4474862f049bf1d8ed4c77/scripts/build-o1js-node-artifacts.sh#L39


Trivo25 (approved):
Thanks!!

PR #303, opened 2024-09-25
handle arrays and tuples 

Trivo25:


mitschabaude (approved):
I think Veridise frowned upon that mix of object types, that's why the error was originally added, but I think this is perfectly fine

PR #310, opened 2024-10-24
Support Secp256r1

mitschabaude:
companion of https://github.com/o1-labs/o1js/pull/1885

* support general curve parameter `a` in most elliptic curve methods
* change many EC methods to take the additional `a` parameter
* support curve parameter `a = -3` in projective curve doubling
  * this is the parameter used by secp256r1
  * note: there is a more efficient algorithm for a=-3 than for general `a`. thanks to the other changes here, it would now be _very_ simple to add the general `a` case as well, but for testing we would need a concrete curve with general `a`, and I'm skipping the work of finding one since I'm not interested in general curves here
* unit-test the new curve (and also its base field)

### DIFF
--- a/crypto/elliptic-curve.ts
+++ b/crypto/elliptic-curve.ts
@@ -402,7 +402,7 @@ function projectiveOnCurve(
   a: bigint
 ) {
   // substitution x -> x/z^2 and y -> y/z^3 gives
-  // the equation y^2 = x^3 + a*z^4 + b*z^6
+  // the equation y^2 = x^3 + a*x*z^4 + b*z^6
   // (note: we allow a restricted set of x,y for z==0; this seems fine)
   let x3 = mod(mod(x * x, p) * x, p);
   let y2 = mod(y * y, p);
--- a/crypto/elliptic-curve.ts
+++ b/crypto/elliptic-curve.ts
@@ -28,7 +28,7 @@ export {
   projectiveToAffine,
   projectiveZero,
   projectiveAdd,
-  projectiveDouble,
+  getProjectiveDouble,
   projectiveNeg,
 };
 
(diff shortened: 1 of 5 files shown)


mitschabaude:
@Trivo25 can you get me the necessary reviewers for this?

Geometer1729 (commented):
LGTM

crypto/elliptic-curve.ts
```diff
@@ -530,14 +623,12 @@ function createCurveAffine({
   endoScalar,
   endoBase,
 }: CurveParams) {
-  // TODO: lift this limitation by using other formulas (in projectiveScale) for a != 0
-  if (a !== 0n) throw Error('createCurveAffine only supports a = 0');
```
querolita, line 534: Removing this check implies you allow any `a`? In `createCurveProjective` you removed the check and called `createProjectiveDouble` where there's a check that `a` is 0 or -3. But what about here?

crypto/elliptic-curve.ts
```diff
@@ -328,20 +395,31 @@ function projectiveEqual(g: GroupProjective, h: GroupProjective, p: bigint) {
...
   // (note: we allow a restricted set of x,y for z==0; this seems fine)
   let x3 = mod(mod(x * x, p) * x, p);
   let y2 = mod(y * y, p);
-  let z3 = mod(mod(z * z, p) * z, p);
-  let z6 = mod(z3 * z3, p);
-  return mod(y2 - x3 - b * z6, p) === 0n;
+  let z2 = mod(z * z, p);
+  let z4 = mod(z2 * z2, p);
+  let z6 = mod(z4 * z2, p);
+  return mod(y2 - x3 - a * x * z4 - b * z6, p) === 0n;
```
querolita, line 412: This last equation does not seem to align with the comment above. Meaning, it says $y^2 = x^3 + az^4 + b z^6$ but the `a` term seems to have an additional `x` here `y2 - x3 - a * x * z4 - b * z6`

querolita (commented):
Left two observations, other than that looks good

crypto/elliptic-curve.ts
mitschabaude: good catch! the comment is wrong :)

crypto/elliptic-curve.ts
mitschabaude: that's a good point, I mean we do support general `a` now in all the purely affine methods like `affineAdd`, `affineDouble`. the only place which doesn't support general a is `projectiveDouble`, which is used here by methods like `scale()` and `isInSubgroup()`

so I think we can accept general `a` here at this point and only when using certain unsupported methods users would hit the error.

Trivo25 (approved):

querolita (approved):

PR #315, opened 2024-11-23
Add support for more branches

Trivo25:
adds support for more branches (branch ^= methods)


querolita (approved):
Interesting, but isn't there a way to allow any amount? Without requiring us to manually increase the Nx used for lists.

hattyhattington17:
I was wondering about these hardcoded modules too - it sounds like we had a discussion around how we could make them dynamic but the associated issue was closed https://github.com/o1-labs/o1js/issues/934. @Trivo25, did we determine that it was not feasible?

hattyhattington17 (approved):

PR #317, opened 2024-11-26
Add parameters of Ed25519 and support twisted Edwards curves

querolita:
This is related to https://github.com/o1-labs/o1js/pull/1283


Trivo25 (commented):
can we remove the .DS_Store file?

querolita:
The files that this PR is directly modifying are:
crypto/elliptic-curve-examples.ts
crypto/elliptic-curve.ts
crypto/elliptic-curve.unit-test.ts

crypto/elliptic-curve-examples.ts
```diff
@@ -55,3 +60,21 @@ const CurveParams = {
   Pallas: pallasParams,
   Vesta: vestaParams,
 };
+
+// https://datatracker.ietf.org/doc/html/rfc8032#section-5.1
+const ed25519Params: TwistedCurveParams = {
+  name: 'Ed25519',
+  modulus: exampleFields.f25519.modulus, // 2^255 - 19
+  order: 0x1000000000000000000000000000000014def9dea2f79cd65812631a5cf5d3edn, //2^252 + 27742317777372353535851937790883648493,
+  cofactor: 8n,
```
querolita, line 69: This is an optional parameter, please a pair of eyes to double check it's correct.

crypto/elliptic-curve.unit-test.ts
```diff
@@ -164,3 +165,73 @@ function curveWithFields(params: CurveParams) {
   // return Curve, Field and Scalar
   return [Projective, Affine.Field, Affine.Scalar] as const;
 }
+
+// Twisted Edwards curve tests
+
+const Ed25519 = createCurveTwisted(TwistedCurveParams.Ed25519);
```
querolita, line 171: What's your preferred convention? CurveTwisted or TwistedCurve? Asking because in affine I believe I have read both.

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+// a * x^2 + y^2 = 1 + d * x^2 * y^2
+// in projective coordinates is represented as
+// a * X^2 * Z^2 + Y^2 Z^2 = Z^4 + d * X^2 * Y^2
+// where x = X/Z, y = Y/Z, and Z ≠ 0
+function projectiveToTwisted(g: GroupProjective, p: bigint): GroupTwisted {
+  let z = g.z;
+  if (z === 0n) {
+    // degenerate case
+    return twistedZero;
+  } else if (z === 1n && g.x === 0n && g.y === 1n) {
```
querolita, line 971: If I didn't treat this case separately, the following `else if` would place a `false` in the `infinity` parameter and then unit tests would complain because they used to find assertions of (0,1,false) != (0,1,true). But note that this is a branch that does not exist in the affine counterpart, and it is assuming that the "only" infinity point is exactly (0,1,1).

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+  generator,
+  a,
+  d,
+}: TwistedCurveParams) {
+  let hasCofactor = cofactor !== undefined && cofactor !== 1n;
+
+  const Field = createField(p);
+  const Scalar = createField(order);
+  const one = { ...generator, infinity: false };
+  const Endo = undefined; // for Ed25519
```
querolita, line 1006: Not sure how to generalize this. When we use this interface for something different than Ed25519, maybe this parameter should change.

querolita (commented):
left some comments for reviewers

crypto/elliptic-curve-examples.ts
dannywillems: LGTM:
```
|     c     | base 2 logarithm of cofactor of edwards25519 in       |
|           | [[RFC7748](https://datatracker.ietf.org/doc/html/rfc7748)] (i.e., 3)
   ```

**Comment by querolita** [comment]
TODO: subgroup check

File: crypto/elliptic-curve.unit-test.ts [hunk 5]
```diff
@@ -164,3 +165,78 @@ function curveWithFields(params: CurveParams) {
   // return Curve, Field and Scalar
   return [Projective, Affine.Field, Affine.Scalar] as const;
 }
+
+// Twisted Edwards curve tests
```
dannywillems, line 169: Can you have some test vectors? I.e. some points defined in the specification that we know are on the curves, in the prime subgroups, and some known results and inputs for each operation?

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+function createCurveTwisted({
+  name,
+  modulus: p,
+  order,
+  cofactor,
+  generator,
+  a,
+  d,
+}: TwistedCurveParams) {
+  let hasCofactor = cofactor !== undefined && cofactor !== 1n;
```
dannywillems, line 1001: Oh boy. Don't tell me that you have to check both equalities in Typescript?!?!?!? :scream: Is it not `cofactor !== 1n` sufficient?

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+};
+
+function twistedOnCurve(
+  { x, y, infinity }: GroupTwisted,
+  p: bigint,
+  a: bigint,
+  d: bigint
+) {
+  if (infinity) return true;
+  // a * x^2 + y^2 = 1 + d * x^2 * y^2
```
dannywillems, line 766: I guess we want to check if it is in the prime subgroup too? If yes, I would create a second method instead of adding in `isOnCurve`.

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+  a: bigint,
+  d: bigint
+): GroupTwisted {
+  if (g.infinity) return h;
+  if (h.infinity) return g;
+
+  let { x: x1, y: y1 } = g;
+  let { x: x2, y: y2 } = h;
+
+  if (y1 === y2) {
```
dannywillems, line 785: LGTM.

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+  d,
+}: TwistedCurveParams) {
+  let hasCofactor = cofactor !== undefined && cofactor !== 1n;
+
+  const Field = createField(p);
+  const Scalar = createField(order);
+  const one = { ...generator, infinity: false };
+  const Endo = undefined; // for Ed25519
+
+  assert(a !== 0n, 'a must not be zero');
```
dannywillems, line 1008: Also, the addition formula works only if d is *not* a square.

dannywillems: Also, `a` should be a square for the addition formula to be complete.

crypto/elliptic-curve.ts
dannywillems: Ah no, ok, it is because you accept it to be undefined for a cofactor of 1. Closing.

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+  let { x: x2, y: y2 } = h;
+
+  if (y1 === y2) {
+    // g + g --> double
+    if (x1 === x2) return twistedDouble(g, p, a, d);
+    // g - g --> return zero
+    if (x1 === mod(p - x2, p)) return twistedZero;
+  }
+
+  // x3 = (x1 * y2 + y1 * x2) / (1 + d * x1 * x2 * y1 * y2)
```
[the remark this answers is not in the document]
dannywillems: LGTM.

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+  d: bigint
+) {
+  let gProj = projectiveFromTwisted(g);
+  let sgProj = projectiveScaleTwisted(gProj, s, p, a, d);
+  return projectiveToTwisted(sgProj, p);
+}
+
+// https://www.hyperelliptic.org/EFD/g1p/auto-twisted-projective.html
+// https://eprint.iacr.org/2008/013.pdf Section 6
+function projectiveAddTwisted(
```
dannywillems, line 873: LGTM

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+  let J = mod(F - 2n * H, p);
+
+  return {
+    x: mod((B - C - D) * J, p),
+    y: mod(F * (E - D), p),
+    z: mod(F * J, p),
+  };
+}
+
+function projectiveScaleTwisted(
```
dannywillems, line 936: LGTM.

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+    return twistedZero;
+  } else if (z === 1n) {
+    // any other normalized affine form
+    return { x: g.x, y: g.y, infinity: false };
+  } else {
+    let zinv = inverse(z, p)!; // we checked for z === 0, so inverse exists
+    // x/z
+    let x = mod(g.x * zinv, p);
+    // y/z
+    let y = mod(g.y * zinv, p);
```
dannywillems, line 982: I would enforce that if x === 0n and y === 1n, then we have the point at infinity. It happens when y = z.
Any point in projective coordinates that is in the representive class of (0 : 1) should be treated the point at infinity.

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+    zero: twistedZero,
+    one,
+
+    hasEndomorphism: Endo !== undefined,
+    get Endo() {
+      if (Endo === undefined) throw Error(`no endomorphism defined on ${name}`);
+      return Endo;
+    },
+
+    from(g: { x: bigint; y: bigint }): GroupTwisted {
```
dannywillems, line 1039: Don't you want to enforce to have x and y given as canonical values? (i.e. smaller than the order of the base field?). I would be highly in favor of this.

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+      if (g.x === 0n && g.y === 1n) return twistedZero;
+      return { ...g, infinity: false };
+    },
+
+    fromNonzero(g: { x: bigint; y: bigint }): GroupTwisted {
```
dannywillems, line 1044: Same comment than above.

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+
+    hasEndomorphism: Endo !== undefined,
+    get Endo() {
+      if (Endo === undefined) throw Error(`no endomorphism defined on ${name}`);
+      return Endo;
+    },
+
+    from(g: { x: bigint; y: bigint }): GroupTwisted {
+      if (g.x === 0n && g.y === 1n) return twistedZero;
+      return { ...g, infinity: false };
```
dannywillems, line 1041: Can you comment that there is no additional check? We should check it is on the curve AND in the prime subgroup.

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+    one,
+
+    hasEndomorphism: Endo !== undefined,
+    get Endo() {
+      if (Endo === undefined) throw Error(`no endomorphism defined on ${name}`);
+      return Endo;
+    },
+
+    from(g: { x: bigint; y: bigint }): GroupTwisted {
+      if (g.x === 0n && g.y === 1n) return twistedZero;
```
dannywillems, line 1040: Note that if g.x and g.y are not enforced to be smaller than the field, the input (0, p + 1) won't give me the identity.

dannywillems: I would add a test for this (i.e. inputs with elements higher than the field. One for each case:
- x < p, y > p
- x > p, y < p
- y > p, x > p.

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+      return { ...g, infinity: false };
+    },
+
+    fromNonzero(g: { x: bigint; y: bigint }): GroupTwisted {
+      if (g.x === 0n && g.y === 1n) {
```
dannywillems, line 1045: See comments online for this. Missing modulus.

crypto/elliptic-curve.unit-test.ts
```diff
@@ -164,3 +165,78 @@ function curveWithFields(params: CurveParams) {
...
+      'not equal to double of itself (or zero)'
+    );
+    assert(
+      !equal(X, negate(X)) || equal(X, zero),
+      'not equal to negation of itself (or zero)'
+    );
+    assert(!equal(X, Y) || X == Y, 'not equal (random points)');
+
+    // algebraic laws - addition
+    assert(equal(add(X, Y), add(Y, X)), 'commutative');
```
dannywillems, line 213: Nice.

crypto/elliptic-curve.unit-test.ts
```diff
@@ -164,3 +165,78 @@ function curveWithFields(params: CurveParams) {
...
+      'distributive'
+    );
+    assert(
+      equal(scale(X, Scalar.negate(x)), negate(scale(X, x))),
+      'distributive (negation)'
+    );
+    assert(
+      equal(scale(X, Scalar.mul(x, y)), scale(scale(X, x), y)),
+      'scale / multiply is associative'
+    );
```
dannywillems, line 240: Can we have some tests with:
- points on curve but not in subgroup
- points not on curve

I want to be sure these inputs raise errors when `from` is used, except if we qualify them as "unsafe constructor". I would then document it.

crypto/elliptic-curve.unit-test.ts
```diff
@@ -164,3 +165,78 @@ function curveWithFields(params: CurveParams) {
...
+test(
+  randomPoint,
+  randomPoint,
+  randomPoint,
+  randomScalar,
+  randomScalar,
+  randomField,
+  (X, Y, Z, x, y) => {
+    // check on curve
+    assert(isOnCurve(X) && isOnCurve(Y) && isOnCurve(Z), 'on curve');
```
dannywillems, line 198: I would like to have negative tests also.

crypto/elliptic-curve.ts
querolita: not sure if that's true... isn't it complete? I thought if d is a square then the curve is basically not safe in cryptography, but that doesn't mean that the operations do not work there...

crypto/elliptic-curve.ts
```diff
@@ -703,3 +716,369 @@ function createCurveAffine({
...
+}
+
+// The twisted curve with equation
+// a * x^2 + y^2 = 1 + d * x^2 * y^2
+// in projective coordinates is represented as
+// a * X^2 * Z^2 + Y^2 Z^2 = Z^4 + d * X^2 * Y^2
+// where x = X/Z, y = Y/Z, and Z ≠ 0
+function projectiveToTwisted(g: GroupProjective, p: bigint): GroupTwisted {
+  let z = g.z;
+  if (z === 0n) {
```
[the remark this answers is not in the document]
querolita: It is a degenerate case, basically this shouldn't occur... In weierstrass you would output the infinity point, but in twisted there's no "infinity" but just the zero point, so it is not the exact same notion either. Let's throw in this case then, good point.

crypto/elliptic-curve.ts
querolita: shouldn't that be already taken into account in the normalized case that z=1, then if x=0 and y=1 the point {0,1} would be returned

crypto/elliptic-curve.ts
querolita: I guess no, and then perform the operations modulo the order and check that it holds (like, {x: 0, y: 1+p} is the identity instead of failing). But I am open to other alternatives, whatever is the standard.

crypto/elliptic-curve.ts
```diff
@@ -703,3 +722,391 @@ function createCurveAffine({
...
+
+  const Field = createField(p);
+  const Scalar = createField(order);
+  const one = { ...generator };
+  const Endo = undefined; // for edwards25519
+
+  assert(a !== 0n, 'a must not be zero');
+  assert(d !== 0n, 'd must not be zero');
+  assert(a !== d, 'a must not be equal to d');
+  // Euler's criterion: d is square iff d^((p - 1) / 2) = 1 mod p
```
Trivo25, line 1041: `iff`

mitschabaude: math jargon 😄 https://en.wikipedia.org/wiki/If_and_only_if

Trivo25: Ohhhhhhhhh

querolita: I meant `iff` indeed, like if and only if 😄

crypto/elliptic-curve.unit-test.ts
```diff
@@ -257,13 +257,25 @@ test(
...
     );
 
-    // subgroup
+    // subgroup and orders
     let pointOrder8: GroupAffineTwisted = { x: 33554441n, y: 33812324n };
+    let pointOrder4: GroupAffineTwisted = { x: 1n, y: 0n };
+    let pointOrder2: GroupAffineTwisted = { x: 0n, y: -1n };
     assert(
-      isOnCurve(pointOrder8) && isInSubgroup(pointOrder8),
+      isOnCurve(pointOrder8) && !isInSubgroup(pointOrder8),
```
dannywillems: Nice.

crypto/elliptic-curve.unit-test.ts
```diff
@@ -257,13 +257,25 @@ test(
...
       'point on curve not in subgroup'
     );
-    assert(equal(scale(pointOrder8, 8n), zero), 'point has order 8');
+    assert(
+      isOnCurve(pointOrder4) && !isInSubgroup(pointOrder4),
```
dannywillems: Nice.

crypto/elliptic-curve.unit-test.ts
```diff
@@ -257,13 +257,25 @@ test(
...
+      'point on curve not in subgroup'
+    );
+    assert(
+      isOnCurve(pointOrder2) && !isInSubgroup(pointOrder2),
```
dannywillems, line 263: Nice.

Trivo25:
This is finished and ported to its own [package](https://github.com/o1-labs-XT/o1js-eddsa) - great job! (keeping the branch alive just in case)

PR #318, opened 2024-11-28
Introduce recursive proofs from within ZkPrograms

mitschabaude:
companion of https://github.com/o1-labs/o1js/pull/1931

## 1. Worker changes

Previously, we properly managed thread pool accesses in the Node.js version using a little ad-hoc state machine. Thanks to that, creating a proof from within a witness block inside another proof (needed in https://github.com/o1-labs/o1js/pull/1931) just worked in Node.js from the get-go.

However, the web version naively tried to launch a second thread pool, which doesn't work because on the Rust side the thread pool is a single global object.

This PR moves the thread pool management into a separate file, porting it to TS in the process, and uses it in the web version as well.

I manually confirmed that nested proofs work in the browser, by running
```sh
./run-in-browser.js src/tests/inductive-proofs-internal.ts 
```
and navigating to the web page that this script serves.

## 2. Pickles bindings changes

A second change here was necessary because the API used to dictate that we have all recursive proofs ready when calling the prover. Since proofs can now be created inside a circuit, I changed the bindings API such that it expects the proofs to be returned from the JS circuit, and not already as direct argument to the prover


js/node/node-backend.js
```diff
@@ -26,59 +26,7 @@ if (!isMainThread) {
 }
 
 // state machine to enable calling multiple functions that need a thread pool at once
-let state = 'none'; // 'initializing', 'running', 'exiting'
```
mitschabaude, line 29: this logic was lifted into `workers.ts` in o1js, see https://github.com/o1-labs/o1js/pull/1931

Trivo25 (approved):

PR #319, opened 2024-12-04
Treat devnet and testnet as the same networkId

hattyhattington17:
Required for https://github.com/o1-labs/o1js/pull/1938


45930:
Are the large diffs related to the primary change?

45930:
image

It's unfortunate since it looks like there is literally no change in those files...

45930 (approved):

mitschabaude:
> It's unfortunate since it looks like there is literally no change in those files...

There is literally no change in those files! When only the ocaml code changes, Wasm files don't need updating.

The annoying thing is that wasm-bindgen will create slightly shuffled versions of its output each time _anything_ changes, even if it doesn't affect the Wasm source.

But the Wasm files could nevertheless be removed from the diff!

mitschabaude:
Actually, I just compiled something with an Ocaml-only change and the wasm files didn't change. @hattyhattington17 you sure you had the correct proof-systems / Mina branches checked out when compiling?

hattyhattington17:
I recloned `o1js` with submodules, checked out this branch within bindings, and then reran the build so I think I have the right code but tell me if I've missed a step! @Trivo25 mentioned that this might be caused by differences in my build environment? I'm using nix on a mac

mitschabaude:
> I recloned `o1js` with submodules, checked out this branch within bindings, and then reran the build so I think I have the right code but tell me if I've missed a step! @Trivo25 mentioned that this might be caused by differences in my build environment? I'm using nix on a mac

Right it's probably the build environment

45930:
Lol, I looked closer at my screenshot and see the diff now.  It's `fP` and `fQ`.  I'm not sure why different builds would reorder those.  Maybe it's non-deterministic?

querolita:
I have just reverted the files modifications causing the ugly diff

PR #336, opened 2025-02-12
New v2 bindings layer

Trivo25:
https://github.com/o1-labs/o1js/pull/2032


mina-transaction/v2/leaves.ts
```diff
@@ -0,0 +1,338 @@
+// NOTE: these leaves are currently backwards compatible with the old encoding format, but the
```
Trivo25, line 1: there's a few TODOs and nice to haves in here, I propose we get back to them in the future given we target an incremental approach with lots of iterations

PR #364, opened 2025-04-16
Update bindings for Mina's `compatible`

mrmr1993:


ocaml/lib/local_ledger.ml
```diff
@@ -223,9 +223,10 @@ let protocol_state_of_json =
     json |> Js.to_string |> Yojson.Safe.from_string
     |> Fields_derivers_zkapps.of_json (Lazy.force deriver)
 
-let apply_zkapp_command_transaction l (txn : Zkapp_command.t)
+let apply_zkapp_command_transaction l (txn : Zkapp_command.Stable.Latest.t)
     (account_creation_fee : string)
     (network_state : Mina_base.Zkapp_precondition.Protocol_state.View.t) =
+  let txn = Zkapp_command.write_all_proofs_to_disk txn in
```
dannywillems, line 229: IIUC, it is mostly to bypass the type checker, but no proofs are actually written on disk.

PR #366, opened 2025-04-22
Bump up bindings for 0.2.89 wasm-bindgen

dannywillems:


dannywillems:
Closing. Old as now we moved to o1js.

